PR #86, opened 2020-05-06
Date column refactor

evansd:
There are two big user-facing changes here:

1. Instead of `include_month` and `include_day` arguments, the recommended style is now to supply a `date_format` argument which accepts `YYYY`, `YYYY-MM` or `YYYY-MM-DD`.

2. Instead of supplying `include_measurement_date` or `include_date_of_match` arguments and having an additional output column automatically generated, the recommended style is now to explicitly create the output column using the new `date_of` function.

For example, instead of:
```py
latest_smoking_code=patients.with_these_clinical_events(
    smoking_codes,
    returning="code",
    include_date_of_match=True,
    include_month=True,
)
```
you would now write:
```py
latest_smoking_code=patients.with_these_clinical_events(
    smoking_codes,
    returning="code",
)
latest_smoking_date=patients.date_of(
    "latest_smoking_code", date_format="YYYY-MM"
)
```

Note that all the old arguments are still supported so existing studies should run without modification.

These user facing changes enable a fair amount of internal simplification as well so there's quite a lot going on in this PR. But the individual commits should hopefully all make sense.


datalab_cohorts/__init__.py
```diff
@@ -1637,19 +1586,48 @@ def with_death_recorded_in_cpns(
...
+        # Deprecated options kept for now for backwards compatibility
+        include_month=False,
+        include_day=False,
     ):
         returning = "date"
-        return "value_from", locals()
+        return "value_from", process_arguments(locals())
+
+
+def process_arguments(args):
```
sebbacon, line 1561: A docstring would be useful here as a hint to future coders about what the expected responsibilities of this function are

datalab_cohorts/__init__.py
```diff
@@ -1637,19 +1586,48 @@ def with_death_recorded_in_cpns(
...
+    validate_time_period_options(**args)
+
+    for date_arg in ("reference_date", "start_date", "end_date"):
+        if date_arg in args:
+            value = args[date_arg]
+            if value == "today":
+                value = datetime.date.today()
+            args[date_arg] = datetime.date.fromisoformat(str(value))
+
+    # Handle deprecated API
```
sebbacon, line 1571: Is there any value in raising a deprecation warning here? I think on balance there probably is; though on the other hand we only have two studies at the moment and could just go and fix them and them remove the compatibility stuff

datalab_cohorts/__init__.py
```diff
@@ -426,16 +456,47 @@ def flatten_nested_covariates(self, covariate_definitions):
...
+
+        The sharp-eyed may notice that this is close to being the inverse of
+        `apply_compatibility_fixes` above and wonder what is being acheived.
+        But this gives us a way of incrementally refactoring things without
+        having to make big sweeping changes and without breaking backwards
+        compatibility.
+        """
+        updated = copy.deepcopy(covariate_definitions)
+        for name, (query_type, query_args) in updated.items():
+            if query_type == "value_from" and query_args["returning"] == "date":
```
sebbacon, line 473: Do we need an assertion here that `date` is the only supported `returning` value?

evansd: At the moment, `value_from` isn't part of the public API, you can only get to it via `date_of`. But you're right, an extra assert can't hurt.

datalab_cohorts/__init__.py
evansd: I'd be more inclined to update the studies and remove the BC code. I included BC because it wasn't much extra work and it makes the whole process much easier, but I'd prefer not to carry it around long term.

evansd:
@sebbacon I'm going to merge this now to avoid building up a stack of dependent PRs. You've already reviewed the bulk of it and everything I've done since is uncontroversial apart from 30793e0 which it might be worth having a quick look at.

PR #95, opened 2020-05-12
Fix data generation for number_of_episodes return type

evansd:


evansd:
Self merging in the interests of time

PR #97, opened 2020-05-12
Remove use of Python 3.8 walrus operator

evansd:
I thought this would be a fun occasion to try it out as we're
controlling the Python version within Docker. But of course `run.py` is
run _outside_ of Docker...


evansd:
Caroline only had 3.7 and this didn't feel like the moment to be making anyone's life more difficult!

PR #99, opened 2020-05-13
New date distributions

alexwalkercebm:


datalab_cohorts/expectation_generators.py
```diff
@@ -40,26 +41,29 @@ def generate_dates(population, earliest_date, latest_date, rate):
...
+        distribution = uniform.rvs(size=int(population)) * elapsed_days
+        distribution = distribution.astype("int")
+    else:
+        print("Only exponential and uniform distributions currently supported")
 
     # And then sample it back down to the requested population size
     distribution = np.random.choice(distribution, population, replace=False)
 
-    df = pd.DataFrame(sorted(distribution), columns=["days"])
+    df = pd.DataFrame(distribution, columns=["days"])
```
sebbacon, line 66: Why remove the sort?

datalab_cohorts/expectation_generators.py
```diff
@@ -40,26 +41,29 @@ def generate_dates(population, earliest_date, latest_date, rate):
...
+        distribution = (
+            expon.rvs(loc=0, scale=0.1, size=int(population * oversample_ratio))
+            * elapsed_days
+        ).astype("int")
+        distribution = distribution[distribution <= elapsed_days]
+    elif rate == "uniform":
+        distribution = uniform.rvs(size=int(population)) * elapsed_days
+        distribution = distribution.astype("int")
+    else:
+        print("Only exponential and uniform distributions currently supported")
```
sebbacon, line 61: I'd explicitly say `exponential_increase` here both (a) because we could imagine decreases in the future and (b) because that's the actual keyword (I realise that the underscore wasn't there previously!)

datalab_cohorts/expectation_generators.py
alexwalkercebm: It meant that the dates weren't randomly distributed, meaning that patients at the top of the [csv](https://gist.github.com/alexwalkercebm/8de0c7ec7c89c3184aae5f8d9ddff5bb) all had very recent outcomes and covariates, while ones lower down had ones from further back. Do we not want them all to be randomly distributed?

datalab_cohorts/expectation_generators.py
evansd, line 61: We should raise an error here, rather than just print a warning e.g:
```
raise ValueError("Only exponential and uniform distributions currently supported")
```

alexwalkercebm: agree, have done

alexwalkercebm: have done

datalab_cohorts/expectation_generators.py
sebbacon: Hm, now you put it like that.... I guess so!

PR #159, opened 2020-06-13
Tidy up packaging and docs in relation to repo rename

sebbacon:
This PR aims to make it easy for people to `pip install` the package.

It also improves Docker image publication so it only happens when the version is bumped.


.github/workflows/build_and_publish.yaml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          fetch-depth: 2
+      - name: Tag
+        id: tag
+        uses: salsify/action-detect-and-tag-new-version@v1
+        with:
+          version-command: |
+            cat runner/VERSION
+
+  build-and-publish-package:
```
inglesp, line 22: Well done for not calling it a gem!

README.md
```diff
@@ -1,17 +1,50 @@
-# OpenSAFELY Research Template
+# OpenSAFELY extractor tool
```
inglesp, line 1: Should this be "OpenSAFELY cohort extractor tool"?

.github/workflows/build_and_publish.yaml
```diff
@@ -0,0 +1,60 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+    - name: Build package
+      run: |
+        python setup.py bdist
```
evansd, line 37: I'll need to double-check this but I _think_ we want `sdist bdist_wheel` here, rather than `bdist`.

setup.py
```diff
@@ -22,4 +22,5 @@
         "console_scripts": ["cohortextractor=datalab_cohorts.cohortextractor:main"]
     },
     include_package_data=True,
+    classifiers=["License :: OSI Approved :: GNU General Public License v3 (GPLv3)"],
```
evansd, line 25: Not disagreeing with this choice, but was there any discussion around this? In the past we've opted for MIT for Datalab stuff on the grounds of maximising permissiveness, but obviously the context is quite different here.

.github/workflows/build_and_publish.yaml
evansd: Yes, it [looks like](https://pythonwheels.com/#pure-wheel) we need to do nothing more than change `bdist` to `sdist bdist_wheel`. If that fails then we need to shove a `pip install wheel` line above, but I have a feeling that's redundant these days.

PR #161, opened 2020-06-15
Add ACME backend

evansd:
Original text by @inglesp 

--------

Here's a draft PR that brings our ACME backend closer to something that can run against ACME's data lake.

I've removed all the covariate functions apart from those used in the risk factors research, and I've marked all test tables that do not match the ACME data lake.  Other test tables do now (more-or-less, see below) match the column names and types in the ACME data lake.

The following covariate functions require ACME to add data to the lake, and as far as I know, this has not happened yet:

* `registered_practice_as_of`, for STP and Region
* `address_as_of`, for IMD and rural/urban
* `admitted_to_icu`
* `with_these_codes_on_death_certificate`
* `died_from_any_cause` 
* `with_death_recorded_in_cpns`

There are some TODOs:

* In the ACME data lake, columns containing SNOMED concept IDs have type `BigInteger`, but our existing machinery expects codes to be strings, and this PR does not change that.  SNOMED IDs are opaque and could be treated as strings.  There are several options:
  * Ask ACME to change the SQL for the materialised views we use, and cast to strings there
  * Create our own materialised views (we may want to do that anyway, for performance -- but this then means we're responsible for updating them)
  * Fix the code, which will mean surgery to eg [`_type_from_return_value`](https://github.com/ebmdatalab/opensafely-research-template/blob/master/datalab_cohorts/process_covariate_definitions.py#L312)
* Some medications don't have SNOMED codes.  I hadn't spotted that until this afternoon, and we'll need to discuss this.
* We need to identify SNOMED codes for BMI

### DIFF
--- a/datalab_cohorts/acme_backend.py
+++ b/datalab_cohorts/acme_backend.py
@@ -317,10 +317,10 @@ class ACMEBackend:
         bmi_cte = f"""
         SELECT t."registration-id", t.BMI, t."effective-date"
         FROM (
-          SELECT "registration-id", NumericValue AS BMI, "effective-date",
+          SELECT "registration-id", "value-pq-1" AS BMI, "effective-date",
           ROW_NUMBER() OVER (PARTITION BY "registration-id" ORDER BY "effective-date" DESC) AS rownum
-          FROM CodedEvent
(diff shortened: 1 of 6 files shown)


inglesp:
Thanks @evansd!

PR #165, opened 2020-06-17
Update safe chars for TPP vaccinations table

evansd:
We need to support such vaccine names as:

    dTaP/IPV
    dTP + Polio
    V419 (PR51)


evansd:
Oh great. The ACME tests failed with:
```
prestodb.exceptions.PrestoQueryError: PrestoQueryError(
    type=INTERNAL_ERROR,
    name=NO_NODES_AVAILABLE,
    message="No nodes available to run query",
    query_id=20200617_104123_00004_s446f
)
```
Which means all that work to ensure that the Presto instance has nodes available before the tests run doesn't actually work.

PR #171, opened 2020-06-17
Only build package/image on push/merge to master

### LINKED ISSUE #168: Only create new tagged versions in master branch
The Github Action which spots changes to the version file and creates new release tag, docker image etc. should only run on the master branch. At present, if you change the version in a PR it creates a new release before the PR is merged. And then after it's merged it generates an error when it tries to create a new tag of the same name.

sebbacon:
Fixes #168

The downside to this is that it makes testing docker images a bit fiddly but I think this is the way round that creates least confusing.

Testing a docker image can still happen by building it from the command line.

### DIFF
--- a/.github/workflows/build_and_publish.yaml
+++ b/.github/workflows/build_and_publish.yaml
@@ -1,6 +1,8 @@
 name: Tag repo; build and publish assets
 on:
   push:
+    branches:
+      - master
 
 jobs:
   tag-new-version:


sebbacon:
I'm awaiting Github's decision on their default branch name.

evansd:
Ah I see, yes: `main` is the presumptive nominee but not yet the candidate.

PR #177, opened 2020-06-17
Support querying household membership

evansd:
Return information about the household to which the patient belonged as of
the reference date. This is inferred from address data using an algorithm
developed by TPP (to be documented soon) so the results are not 100%
reliable but are apparently pretty good.

Options for `returning` are:

    pseudo_id: An integer identifier for the household which has no meaning
               other than to identify individual members of the same
               household (0 if no household information available)

    household_size: the number of individuals in the household (0 if no
                    household information available)

Examples:

    household_id=patients.household_as_of(
        "2020-02-01", returning="pseudo_id"
    )

    household_size=patients.household_as_of(
        "2020-02-01", returning="household_size"
    ),


sebbacon:
Great. I know this is a bit "pending a better solution" but can we add the documentation to https://github.com/opensafely/documentation/blob/master/study_definition.md for the time being?

PR #178, opened 2020-06-17
Fix SNOMED code column type

inglesp:
With a tiny bit of special-casing in `get_column_expression`, this turns out not to be that much of a change.


inglesp:
Hi @evansd, thanks for this!  Could you re-review?

inglesp:
Changing all the codes to strings in the tests picked up something else to be fixed in `codelist_to_sql`.

Another review please @evansd!

PR #180, opened 2020-06-18
Add crude (but hopefully effective) error checking

### LINKED ISSUE #164: Error checking when no DATABASE_URL defined
At the moment we rely on the backend init process to do most of the error checking of a study definition. But now we support using a study definition to generate dummy data without defining a backend at all which means you can write invalid study defs and still get data from them (see https://github.com/opensafely/research-template/issues/12).

We should fix this. Somehow.

evansd:
This fixes #164 (at least for now). We'll have to consider how this
interacts with the ACME backend once we start using this in earnest, but
that's for another day.


tests/test_study_definition.py
```diff
@@ -54,3 +54,13 @@ def test_unrecognised_database_url_raises_error(tmp_path, monkeypatch):
             sex=patients.sex(),
             age=patients.age_as_of("2020-01-01",),
         )
+
+
+def test_errors_are_triggered_without_database_url(tmp_path, monkeypatch):
+    monkeypatch.delenv("DATABASE_URL", raising=False)
+    with pytest.raises(Exception):
```
inglesp, line 61: Can we check that the expected exception is raised?

evansd: Ah yes, fair point.

PR #196, opened 2020-06-23
Ensure SQL errors are propagated

### LINKED ISSUE #189: SQL-level errors do not propagate
I just managed to generate a cohort in the Extract database which contained not data, but a SQL Server error message related to a missing (`Msg 208, Level 16, State1, ` ... `Invalid object name`) table.

This should have resulted in an error.

sebbacon:
Using `sqlcmd` in batch mode causes errors to be printed to stdout
without any error status

Fixes #189


sebbacon:
Not this kind of error, no

Issue #182, opened 2020-06-19
Add ECDS data to backend


inglesp:
Split out from https://github.com/opensafely/cohort-extractor/issues/163.

> We want:
> 
> * Date of attendance
> * No of attendances in X period of time
> * COVID attendance (from codelist)
> * Discharge destination

inglesp:
Proposed API:

```
# date of attendance
patients.admitted_to_emergency_care(
    on_or_after="2020-02-01",
    returning="date_admitted"  # or "binary_flag"
)

# no of attendances in in period of time
patients.admitted_to_emergency_care(
    between=["2020-02-01", "2021-02-01"],
    returning="number_of_matches_in_period"
)

# covid attendance
patients.admitted_to_emergency_care(
    on_or_after="2020-02-01",
    with_these_codes=[...],
    returning="date_admitted"  # or "binary_flag"
)

# discharge destination
patients.admitted_to_emergency_care(
    on_or_after="2020-02-01",
    returning="discharge_destination"
)
```

@CarolineMorton, @alexwalkercebm, does this look like it captures everything you might want to do with ECDS data?

@evansd, @sebbacon, does this fit in with existing patterns?  Does anything look unimplementable?

CarolineMorton:
Yes this looks exactly what we need. Thanks @inglesp

HelenCEBM:
Some notes on this here: https://github.com/opensafely/outcomes-notebook/issues/3

Some points and questions:
- I wouldn't use the phrase `admitted_to_emergency_care` because that usually means being admitted to a ward/bed from A&E and so may cause confusion with inpatient records - would suggest `attended_accident_and_emergency` or similar is more appropriate.
- Just to check, can the above items be combined? E.g. Patient had a covid attendance and was admitted from that particular attendance?
- do we want to have a separate "likely covid" category based upon "shortness of breath" and similar diagnoses? Or will that be built into the "covid attendance" item in the study def as needed?
- a data quality flag might be helpful, e.g. if a trust does not complete diagnosis codes then the flag for covid attendance cannot be completed. 
- I doubt we will need all the possible outputs from discharge_destination - there are a dozen or so but this could be simplified to a flag for "admitted/transferred" (vs "not admitted")

inglesp:
[...]

PR #205, opened 2020-06-23
Add ECDS to TPP backend

inglesp:
Addresses #182.


sebbacon:
`discharge_destination` should perhaps be a codelist?

cohortextractor/patients.py
```diff
@@ -530,3 +530,31 @@ def household_as_of(reference_date, returning=None, return_expectations=None):
...
+    with_these_diagnoses=None,
+    discharged_to=None,
+    return_expectations=None,
+):
+    """Return information about attendance of A&E.
+
+    Options for `returning` are:
+
+        binary_flag: Whether patient attended A&E
+        date_arrived: date patient arrived in A&E
```
sebbacon, line 552: I see we have `date_admitted` (for ICU api) as a `returning` option already - should we use that name?

inglesp: Helen says that "admitted" might imply that a patient was admitted to hospital from A&E, so I wanted to avoid that term.

sebbacon: OK!

Should we include a summary of the discussions from the issue in the docstring?  (At least link to them I guess)

inglesp: Good idea, I'll update.

PR #207, opened 2020-06-24
Don't require pyodbc for dummy data generation

evansd:
Requiring pyodbc complicates local installation significantly and
there's no reason we actually need it. However my recent "nifty hack" of
using the TPPBackend for error checking ended up importing it.

This PR removes that import and adds a test so I don't make the same
mistake again.


tests/test_study_definition.py
```diff
@@ -56,11 +60,40 @@ def test_unrecognised_database_url_raises_error(tmp_path, monkeypatch):
...
+                    "category": {"ratios": {"M": 0.49, "F": 0.51}},
+                }
+            ),
+        )
+        study.to_csv("/dev/null", expectations_population=10)
+        status = "is" if "pyodbc" in sys.modules else "is not"
+        print(f"pyodbc {status} imported")
+
+    # Convert the code in `test_script` to a standalone script by removing the
+    # function definition line and stripping the indentation
```
inglesp, line 95: Why not put the code in a string?

evansd: Just because I like having it Blackened, linted etc and it was fairly easy to do

PR #210, opened 2020-06-25
Fix definition dumping

sebbacon:


setup.py
```diff
@@ -12,6 +12,7 @@
     url="https://github.com/opensafely/cohort-extractor",
     author="OpenSAFELY",
     author_email="tech@opensafely.org",
+    python_requires=">=3.7",
```
evansd, line 15: We should run the tests against Python 3.7 as well if we're going to say we support it (currently we just test 3.8).

PR #211, opened 2020-06-25
Add backend selector

sebbacon:
Add switches for selecting backends and database flavours.

This still requires support in the job server and the job runner to work.


cohortextractor/remotejobs.py
```diff
@@ -73,8 +73,21 @@ def do_post(data):
...
-def submit_job(tag, operation):
+def submit_job(backend, db, tag, operation):
     allowed_operations = ["generate_cohort"]
+    allowed_backends = ["all", "tpp"]
     assert operation in allowed_operations, f"operation must be in {allowed_operations}"
-    data = {"repo": get_repo(), "tag": tag, "operation": "generate_cohort"}
+    assert backend in allowed_backends, f"backend must be in {allowed_backends}"
+    if backend == "all":
+        backends = allowed_backends[:]
+        backends.remove("all")
```
inglesp, line 83: `backends` won't be defined if `backend` is not `"all"`.

cohortextractor/remotejobs.py
```diff
@@ -73,8 +73,21 @@ def do_post(data):
...
+    repo = get_repo()
+    for backend in backends:
+        data = {
+            "repo": repo,
+            "tag": tag,
+            "operation": "generate_cohort",
+            "backend": backend,
+            "db": db,
+        }
```
inglesp, line 92: `data` will be take whatever value corresponds to the last item of `backends`.

PR #215, opened 2020-06-26
Add dummy data generation to vaccinations study def

evansd:
This can certainly be improved in many ways, but it's better than nothing.


evansd:
It's funny, I really like it because it's how I've always wanted dicts to behave and it simplifies so many things.

There was definitely a time when it was a bit dangerous because you could write code that appeared to run fine on older versions of Python but would be subtlety broken. But given that dicts have been de facto ordered since 3.5 I think we're passed that now.

PR #216, opened 2020-06-26
Enforce a maximum as well as a minimum date of birth

evansd:
It turns out there are some patients in the TPP database with dates of
birth in the future (sometimes by quite a long way). This is infrequent
enough that it's probably ignorable, but it would be better to exclude
them anyway.

At the same time we can exclude patients who are below the age of one
because the youngest patients we will use in our analysis are patients
who turn one this month, and on principle we shouldn't retreive data
we're not going to use.

Note this pulls in commit from #215 which should be merged first, but it
was distinct enough that it deserved its own PR.


evansd:
Yeah, that's a reasonable point. "Scheduled Vaccinations" would be another option as that's what the code's specialised to rather than anything to do with age. But "Childhood Vaccinations" is certainly accurate and probably clearer.

evansd:
An interesting data point: this PR seems to have consistently triggered the test flakiness whereby Presto is not properly initialised when the first test runs. (By "consistent" I mean that I've forced pushed several times and also manually triggered a test re-run and on every occasion the same failure has occurred.)

PR #218, opened 2020-06-30
Add was_ventilated as possible returning value for admitted_to_icu

inglesp:
See https://ebmdatalab.slack.com/archives/C63UXGB8E/p1593514641061200 for discussion

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -274,6 +274,15 @@ def admitted_to_icu(
     include_month=True,
     include_day=False,
 ):
+    """Return information about being admitted to ICU.
+
+    Options for `returning` are:
+
+        binary_flag: Whether patient attended A&E
+        was_ventilated: Whether patient was ventilated
+        date_admitted: Date patient arrived in A&E
+    """
+
     return "admitted_to_icu", locals()
 
 
--- a/cohortextractor/process_covariate_definitions.py
+++ b/cohortextractor/process_covariate_definitions.py
@@ -332,6 +332,7 @@ class GetColumnType:
             "household_size": "int",
             "date_arrived": "date",
             "discharge_destination": "str",
+            "was_ventilated": "bool",
         }
         try:
             return mapping[returning]
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -956,6 +956,9 @@ class TPPBackend:
         elif returning == "binary_flag":
             column_name = "was_admitted"
             column_definition = 1
+        elif returning == "was_ventilated":
+            column_name = "ventilated"
+            column_definition = "MAX(Ventilator)"  # apparently can be 0, 1 or NULL
         else:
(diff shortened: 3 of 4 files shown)


inglesp:
> Would this perhaps be better implemented as an input filter to the API? So you can get the dates for people who were ventilated? I'm not sure... just a question

Are you suggesting a new method on `patients` called `ventilated_in_icu`?

Something like:

```
patients.ventilated_in_icu(
    on_or_after="2020-02-01",
    returning="binary_flag",  # or "date_ventilated"
)
```

?

@CarolineMorton, @rohinimathur, does the new code as used by [this test](https://github.com/opensafely/cohort-extractor/blob/2dcdf7addebb0abbe9bbd5619db5d6a5baec1bec/tests/test_tpp_backend.py#L1280-L1285) look like it does enough?

sebbacon:
More analogue to `with_these_codes`. Something like 

```python
patients.admitted_to_icu(
      on_or_after="2020-02-01",
      returning="binary_flag",
      with_these_characteristics(["ventilated"])
)
```

But I've not given this much thought, in a hurry, so might be a dumb suggestion!

CarolineMorton:
Yes i think this is fine @inglesp 

> Something like:
> 
> ```
> patients.ventilated_in_icu(
>     on_or_after="2020-02-01",
>     returning="binary_flag",  # or "date_ventilated"
> )
> ```

would do the job nicely for @rohinimathur study. A further thing has just come up with @hjforbes study which may be needed in other ones as well. Can we tell what the admitting diagnosis was - ie. was it covid? - from the itu data. Would that be able to be added in?

If that is the case, then @sebbacon approach may make more sense as it would be another characteristic i think?

CarolineMorton:
Another option is to infer covid itu admission from other sources like A&E attendance with covid, positive SGSS tests etc

rohinimathur:
This would be perfect for ethnicity study - then we would get date
b) date of first ventilation event in an ICU admission
c) ventilation yes/no during ICU admission

Do we have a codelist for 'COVID related ICU admissions'? 
Do we need this if we want to get at the outcome of 'Ventilation for COVID related ICU admission'

CarolineMorton:
So on examination of slack archives, it looks like we are only getting covid-positive patients in this dataset so we can just ignore my previous suggestion and go for binary of `ventilation`.

rohinimathur:
Great, so sounds like we have everything we need. Good to note that 'if' ICNARC data are updated in the future to get all ICU admissions rather than just COVID only, then we'll need to refine this definition.

inglesp:
OK, thanks for the discussion!  @rohinimathur sounds like you can go ahead and pretend that

```
patients.ventilated_in_icu(
    on_or_after="2020-02-01",
    returning="binary_flag",  # or "date_ventilated"
)
```

does the right thing, and then when @evansd's back tomorrow we'll confirm that the implementation is correct and then you can start using it.

PR #223, opened 2020-07-01
Switch to Python 3.7 for ease of local installation

### LINKED ISSUE #222: Change to requiring python 3.7
Python 3.8 is not yet supported by Anaconda, which is our currently-recommended installation route for Windows users.

Just to note that at a minimum, this means changing the tests. Then anything else that breaks.

https://github.com/opensafely/documentation/pull/4#pullrequestreview-440099415

evansd:
In particular, Anaconda doesn't currently have good 3.8 support.

As we weren't using any 3.8-isms this doesn't require any code changes.

Closes #222


evansd:
Yes, we should. But I couldn't be bothered to look up Github's job matrix syntax at this point.

evansd:
Made an issue: https://github.com/opensafely/cohort-extractor/issues/224

PR #226, opened 2020-07-01
Fix flaky Presto tests

### LINKED ISSUE #197: Flaky tests against Presto
We often see the first of `test_acme_backend` tests fail because Presto is not ready.

Next time we see this, we'll add the traceback here.

Not urgent to fix, but unhelpful and aesthetically unpleasing.

evansd:
It turns out that Presto will tell you it has active nodes in its `system.runtime.nodes` table but still throw a "no nodes available" error if you try to execute a query which needs to touch the MSSQL instance. This PR works around that issue.

Closes #197

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -35,7 +35,7 @@ exec docker-compose run \
   --volume "$PWD:/workspace" \
   -e PYENV_VERSION=3.7.8 \
   -e TPP_DATABASE_URL='mssql://SA:Your_password123!@mssql:1433/Test_OpenCorona' \
-  -e EMIS_DATABASE_URL=presto://presto:8080/mssql/dbo \
-  -e EMIS_DATASOURCE_DATABASE_URL='mssql://SA:Your_password123!@mssql:1433/Test_EMIS' \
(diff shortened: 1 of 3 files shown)


run.sh
```diff
@@ -35,7 +35,7 @@ exec docker-compose run \
   --volume "$PWD:/workspace" \
   -e PYENV_VERSION=3.7.8 \
   -e TPP_DATABASE_URL='mssql://SA:Your_password123!@mssql:1433/Test_OpenCorona' \
-  -e EMIS_DATABASE_URL=presto://presto:8080/mssql/dbo \
-  -e EMIS_DATASOURCE_DATABASE_URL='mssql://SA:Your_password123!@mssql:1433/Test_EMIS' \
+  -e ACME_DATABASE_URL=presto://presto:8080/mssql/dbo \
+  -e ACME_DATASOURCE_DATABASE_URL='mssql://SA:Your_password123!@mssql:1433/Test_ACME' \
```
inglesp, line 39: :scream:

PR #228, opened 2020-07-03
Raise a meaningful error when MS SQL credentials are wrong

sebbacon:


tests/test_tpp_backend.py
```diff
@@ -48,6 +48,13 @@ def set_database_url(monkeypatch):
         monkeypatch.setenv("DATABASE_URL", os.environ["TPP_DATABASE_URL"])
 
 
+@pytest.fixture
```
evansd, line 51: It seems more complicated than necessary to make this a fixture as I doubt it will be reused. I think you can just make the test function accept `monkeypatch` and then fiddle with DATABASE_URL there.

tests/test_tpp_backend.py
```diff
@@ -103,6 +110,16 @@ def test_sql_error_propagates_with_sqlcmd():
             assert "Invalid object name 'Bar'" in str(excinfo.value)
 
 
+def test_credentials_error_propagates_with_sqlcmd(set_database_url_with_bad_password):
+    with patch.object(TPPBackend, "to_sql") as to_sql:
```
evansd, line 114: Rather than patching, can't we just call `mssql_query_to_csv_file` directly and assert that it raised the right error? You wouldn't even need to monkeypatch DATABASE_URL in that case, as it's supplied directly as an argument.

tests/test_tpp_backend.py
sebbacon: Yes, you're right, the way I did this is left over from an approach I threw away.

PR #234, opened 2020-07-13
feat: Support returning de-registrations dates

evansd:
Normally we require that patients stay registered throughout the study
period, but in certain cases where we have data from outside primary
care we may want to include de-registered patients in a study. For such
patients we need to know the date they de-registered so we can treat
them appropriately in the analysis.

### DIFF
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -848,6 +848,36 @@ class TPPBackend:
             """,
         )
 
+    def patients_date_deregistered_from_all_supported_practices(self, between):
+        if between is None:
+            between = (None, None)
+        min_date, max_date = between
+        # Registrations with no end date are recorded using an end date of
+        # 9999-12-31, so if no max date is supplied then we need to set a max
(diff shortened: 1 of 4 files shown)


cohortextractor/tpp_backend.py
```diff
@@ -848,6 +848,36 @@ def patients_registered_practice_as_of(self, date, returning=None):
...
+    def patients_date_deregistered_from_all_supported_practices(self, between):
+        if between is None:
+            between = (None, None)
+        min_date, max_date = between
+        # Registrations with no end date are recorded using an end date of
+        # 9999-12-31, so if no max date is supplied then we need to set a max
+        # date to something less than 9999-12-31 to filter out registrations
+        # which have no end date.
+        if max_date is None:
+            max_date = "3000-01-01"
```
inglesp, line 860: Giving the Y3K consultants something to do?

evansd: 64KB ought to be enough for anybody...

PR #238, opened 2020-07-15
fix: Support `return_expectations` for de-registrations

evansd:
We need to find some way of making sure we (or I) don't forget to test
for this.


evansd:
Self-merging in the interests of time as it's a tiny fix

PR #240, opened 2020-07-16
fix: Refactor sqlcmd wrapper for better error handling

### LINKED ISSUE #239: Error handling issues with sqlcmd
We had an issue (#236) recently where a divide-by-zero error wasn't immediately obvious because of a cascade of other issues.

One problem is that the output CSV starts with two warning lines before the headers:
screenshot1

(Sub-question: why are we getting these warnings?)

 I think this is something to do with the fact that we only filter errors/warning that happen at the very start of the output:
https://github.com/opensafely/cohort-extractor/blob/96db6c7cc4607887ea0b3218c7606d318e78904e/cohortextractor/mssql_utils.py#L106-L118

We then get a smallish number of output rows before we hit a divide-by-zero error and the output terminates:
screenshot2

Finally, the fact that the same warnings are repeated then causes the duplicate checker to think we have duplicate patient IDs, so it ends up throwing:
```
RuntimeError: Duplicate IDs found (1 rows)
```

I suspect we need just need to extend the error checking logic to cover all lines, but there may well be some reason we didn't do this in the first place.

evansd:
Previously this code used the `-o output_file` option. This caused both
query output and warnings/errors to be mixed together in a single file,
which then required hairy and error prone code to try to separate back
out.

We now just let sqlcmd write to stdout, which we redirect to a file
ourselves, leaving error output written to stderr.

Closes #239

### DIFF
--- a/cohortextractor/mssql_utils.py
+++ b/cohortextractor/mssql_utils.py
@@ -1,4 +1,5 @@
 import os
+import re
 import subprocess
 import tempfile
 from urllib.parse import urlparse, unquote
@@ -105,7 +106,13 @@ def _mssql_query_to_csv_file(database_url, query, filename):
                 # sqlcmd outputs a separator line (consisting of repeated
                 # dashes) between the column headers and the data. There's no
                 # option to disable this behaviour so we remove it here.
-                if line_num == 1 and line.startswith("-"):
+                if line_num == 1:
+                    if not re.match(r"^[\-,]+$", line):
+                        stderr += (
+                            f"\nExpected line 2 to be a separator containing only "
+                            f"dashes and commas but found: {line}"
+                        )
+                        has_error = True
                     continue
                 yield line
                 out.write(line)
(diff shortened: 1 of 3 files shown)


cohortextractor/mssql_utils.py
```diff
@@ -91,33 +90,31 @@ def _mssql_query_to_csv_file(database_url, query, filename):
...
+            _, stderr = process.communicate()
+            has_error = process.returncode != 0 or _output_contains_errors(stderr)
+        # We use windows line endings because that's what the CSV module's
+        # default dialect does
+        with open(filename, "w+", newline="\r\n") as out, open(temp_file, "r") as inp:
+            for line_num, line in enumerate(inp):
+                # sqlcmd outputs a separator line (consisting of repeated
+                # dashes) between the column headers and the data. There's no
+                # option to disable this behaviour so we remove it here.
+                if line_num == 1 and line.startswith("-"):
```
inglesp, line 108: What if `line_num == 1 and not line.startswith("-")`?  Should that blow up?

evansd: That's a good point. I think it should probably continue and then raise an error at the end, as it does with other types of failure. I'll do that now.

PR #245, opened 2020-07-21
feat: add cause of death as return value to ONS death filters

### LINKED ISSUE #244: get cause of death ICD-10 codes into study cohort
We can currently query cause of death codes with `patients.with_these_codes_on_death_certificate` but can't return the codes themselves. Returning the primary cause of death code would be good enough for now. 

Some way to return all cause of death codes might be useful in future (though identifiability issues abound)

sebbacon:
Fixes #244

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -302,6 +302,16 @@ def with_these_codes_on_death_certificate(
     include_day=False,
     return_expectations=None,
 ):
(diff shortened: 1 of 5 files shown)


sebbacon:
@wjchulme this should be reviewed and deployed by tomorrow. Note in the diffs there's some added documentation - might want to include that in the doc repo? (Can't remember where we're at with that)

PR #255, opened 2020-07-30
Add "index_date" support

evansd:
This PR adds "index_date" support to study definitions. Here is a simple example of what this looks like:

```py
study = StudyDefinition(
    index_date="2015-06-01",
    population=patients.with_these_clinical_events(
        copd_codes,
        between=[
            "first_day_of_year(index_date) - 2 years",
            "last_day_of_year(index_date)",
        ],
    ),
    age=patients.age_as_of("index_date"),
)
```

If you define an `index_date` on a study definition then everywhere that you might normally supply a date you can now supply a "date expression".

The simplest date expression is just the string `index_date`, which gets replaced by whatever the index date is set to.

It's also possible to apply various functions to the index date. The available options (hopefully self-explanatory) are:
```
first_day_of_month(index_date)
last_day_of_month(index_date)
first_day_of_year(index_date)
last_day_of_year(index_date) 
```

Finally, intervals of time can be added or subtracted from the index date (or from a function applied to the index date). The available units are `year(s)`, `month(s)` and `day(s)`. For example:
```
index_date + 90 days
first_day_of_month(index_date) + 9 months
index_date - 1 year
```

Note that if the index date is 29 February and you add or subtract some number of years which doesn't lead to a leap year, then an error will be thrown. Similarly, if adding or subtracting months leads to a month with no equivalent day e.g. adding 1 month to 31 January to produce 31 February.

### DIFF
--- a/cohortextractor/date_expressions.py
+++ b/cohortextractor/date_expressions.py
@@ -1,0 +1,110 @@
+import calendar
+import datetime
+import re
+
+
+class InvalidExpressionError(ValueError):
+    pass
+
+
+class UnparseableExpressionError(InvalidExpressionError):
+    pass
+
+
+def create_regex():
+    token = r"[A-Za-z_\-\.]+"
+    name = f"(?P<name>{token})"
+    function = f"(?P<function>{token})"
+    operator = r"(?P<operator> \+ | \-)"
+    quantity = r"(?P<quantity>\d+)"
+    units = f"(?P<units>{token})"
+    return re.compile(
+        rf"^( {function} \( )? {name} \)? ( {operator} {quantity} {units} )?$",
+        re.VERBOSE,
+    )
+
+
+class DateExpressionEvaluator:
+
+    regex = create_regex()
+
+    def __init__(self, index_date):
+        self.index_date = index_date
+
+    def __call__(self, expression_str):
+        match = self.regex.match(expression_str.replace(" ", ""))
+        if not match:
+            raise UnparseableExpressionError(expression_str)
+        try:
+            return self.evaluate(**match.groupdict())
+        except InvalidExpressionError as e:
+            message = f"{e} in: {expression_str}"
+            e.args = (message, *e.args[1:])
+            raise e
+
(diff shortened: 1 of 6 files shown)


cohortextractor/process_covariate_definitions.py
```diff
@@ -382,6 +380,60 @@ def type_of_aggregate_of(self, column_names, aggregate_function, **kwargs):
...
+            )
+        if "return_expectations" in query_args:
+            query_args["return_expectations"] = process_expectations_definition(
+                query_args["return_expectations"], index_date
+            )
+        output[name] = (query_type, query_args)
+    return output
+
+
+def process_expectations_definition(expectations_definition, index_date):
```
sebbacon, line 403: I wonder if this should be `process_date_expressions_in_expectations_definition` or similar, to make it clearer what's happening when it's called?

cohortextractor/process_covariate_definitions.py
```diff
@@ -382,6 +380,60 @@ def type_of_aggregate_of(self, column_names, aggregate_function, **kwargs):
         return column_type
 
 
+def process_date_expressions(covariate_definitions, index_date):
```
sebbacon, line 383: I would find a short docstring helpful here to explain what "process" means

cohortextractor/date_expressions.py
```diff
@@ -0,0 +1,110 @@
...
+    def date_function_last_day_of_year(self, date):
+        return date.replace(month=12, day=31)
+
+    def date_unit_years(self, date, value):
+        # This can potentially throw an error if used on 29 Feb
+        return date.replace(year=date.year + value)
+
+    def date_unit_months(self, date, value):
+        # Can potentially thrown an error if day is greater than 28 and there's
+        # no corresponding day in the resulting month
```
inglesp, line 98: Is it worth catching this?  The error (`ValueError: day is out of range for month`) could be more helpful (eg `3 months after 2020-01-31 doesn't exist`)

PR #269, opened 2020-08-13
Add measure support

evansd:
This PR adds supporting for generating OpenPrescribing syle measures from OpenSAFELY data.

Generating measures is 3 step process:

 1. Defining a `measures` variable in `study_definition.py`, which should be a list of `Measure` instances.
 2. Extracting the data by running `generate_cohort` using the new `--index-date-range` option to cover the range of time periods we want to calculate the measure for.
 3. Running `generate_measures` which takes the files extracted in step 2 and produces files like `measure_<measure_id>.csv`.

## 1. Defining measures

This is best explained using an example:
```py
from cohortextractor import StudyDefinition, Measure, patients


study = StudyDefinition(
    index_date="2020-01-01",
    population=patients.registered_with_one_practice_between(
        "index_date", "index_date + 1 month"
    ),
    admitted_to_icu=patients.admitted_to_icu(
        between=["index_date", "index_date + 1 month"]
    ),
    died=patients.died_from_any_cause(
        between=["index_date", "index_date + 1 month"]
    ),
    stp=patients.registered_practice_as_of(
        "index_date", returning="stp_code",
    ),
)

measures = [
    Measure(
        id="icu_admission_by_stp",
        numerator="admitted_to_icu",
        denominator="population",
        group_by="stp",
    ),
    Measure(
        id="death_by_stp",
        numerator="died",
        denominator="population",
        group_by="stp",
    ),
]
```
Here `id` is just a string used to identify the measure output file. The other attributes identify columns from the study definition used to calculate the measure. Both the `numerator` and `denominator` columns must be numeric (which for our purposes includes booleans, encoded as 0 or 1). The `group_by` column can be of any type. 

## 2. Extracting data

The `generate_cohort` sub-command now accepts an `--index-date-range` option. Rather than generating a single output CSV file this generates multiple output files across a range of dates, modifying the study's index date each time.

The range is specified as:

    --index-date-range 'YYYY-MM-DD to YYYY-MM-DD by (week|month)'

We also except `today` in place of a date.

Output files are named like: `output/input_YYYY-MM-DD.csv`

We also add `--skip-existing` option which will cause the cohortextractor to skip the extraction step for any dates where the corresponding file already exists. This makes it easier to incrementally extract data for new months/weeks without having to re-extract everything.

Example:

    cohortextractor generate_cohort --index-date-range '2020-01-01 to 2020-03-01 by month'

    output/input_2020-01-01.csv
    output/input_2020-02-01.csv
    output/input_2020-03-01.csv

## 3. Generating measures

This involves a new `generate_measures` sub-command:

    cohortextractor generate_measures

For each defined measure, and for each monthly/weekly file extracted in step 2, this generates an output file with the measure calculated for that month or week.

    output/measure_icu_admission_by_stp_2020-01-01.csv
    output/measure_death_by_stp_2020-01-01.csv
    output/measure_icu_admission_by_stp_2020-02-01.csv
    output/measure_death_by_stp_2020-02-01.csv
    output/measure_icu_admission_by_stp_2020-03-01.csv
    output/measure_death_by_stp_2020-03-01.csv

Finally, for each measure, it combines all the output into a single file with an additional `date` column indicating the date associated with each row.

    output/measure_icu_admission_by_stp.csv
    output/measure_death_by_stp.csv

This command also respects the `--skip-existing` flag. This will prevent it from recalculating the measure for any months or weeks which have already been calculated. However the final step, which combines output across time periods, will always be run.

------------------------------------------------

## Implementation notes

At the moment, the structure of the measures is deliberately minimal. We've discussed titles, descriptions, tags etc which are all good ideas. But at present there's nothing we can do with any of this metadata. I'd rather add this in when it actually has a role e.g. when we add chart generation or something like that.

There's a big fat integration test which checks that this all runs and produces sensible output, but there's obvious scope for more unit testing around some of this. However the actual measure calculation itself is three lines of primary school Pandas, so I'm not too worried about that bit.

Probably easiest reviewed by commit.


cohortextractor/cohortextractor.py
```diff
@@ -220,6 +223,78 @@ def _increment_date(date, period):
...
+    """
+    Given a CSV file name and a list of measures, load the file into a Pandas
+    dataframe with types as appropriate for the supplied measures
+    """
+    numeric_columns = set()
+    other_columns = set()
+    for measure in measures:
+        numeric_columns.update([measure.numerator, measure.denominator])
+        other_columns.add(measure.group_by)
+    # This is a special column which we don't load form the CSV but whose value
```
inglesp, line 295: s/form/from/

cohortextractor/cohortextractor.py
```diff
@@ -117,23 +119,92 @@ def generate_cohort(output_dir, expectations_population, selected_study_name=Non
...
+        start = end = date_range_str
+    try:
+        start = _parse_date(start)
+        end = _parse_date(end)
+    except ValueError:
+        raise ValueError(
+            f"Invalid date range '{date_range_str}': Dates must be in YYYY-MM-DD "
+            f"format or 'today' and ranges must be in the form "
+            f"'DATE to DATE by (week|month)'"
+        )
```
sebbacon, line 205: nit: slightly better to validate `week|month` here rather than when it's attempted to be used in `_increment_date`?

cohortextractor/cohortextractor.py
```diff
@@ -371,6 +442,12 @@ def main():
         type=str,
         default=os.environ.get("TEMP_DATABASE_NAME", ""),
     )
+    generate_cohort_parser.add_argument(
+        "--index-date-range",
```
sebbacon, line 571: Thinking aloud: we've talked about making `index-date` a compulsory argument, which I like & think we should implement soon. When we do, would this be a mutually-exclusive option group of either `--index-date` OR `--index-date-range`?  I think that sounds right.

cohortextractor/cohortextractor.py
```diff
@@ -220,6 +223,78 @@ def _increment_date(date, period):
...
+    """
+    numeric_columns = set()
+    other_columns = set()
+    for measure in measures:
+        numeric_columns.update([measure.numerator, measure.denominator])
+        other_columns.add(measure.group_by)
+    # This is a special column which we don't load form the CSV but whose value
+    # is always set to 1 for every row
+    numeric_columns.discard("population")
+    dtype = {col: "str" for col in other_columns}
```
sebbacon, line 298: Given `other_columns` is always the `group_by` (right? in which case, in passing, should the variable name be `group_by_columns`?) and therefore *should* be of low cardinality, I think the appropriate dtype is probably `categorical` - would save lots of memory.

sebbacon (changes requested):

sebbacon:
Any thoughts about real-world performance, by the way? Is it worth trying this against real data before launching it to everyone?

cohortextractor/cohortextractor.py
evansd: I'm not 100% sure about that. I feel like, for a normal study (as opposed to a measure definition) the index date belongs with the rest of the study definition rather than as something supplied on the command line.

I think we would want to encourage people to make use of it, rather than hardcoding dates throughout the study, but that's a slightly different issue.

evansd:
I think for performance we'll just have to suck it and see. We could "soft launch" by merging it but only telling a few people for the purposes of writing some initial experimental measures?

cohortextractor/cohortextractor.py
sebbacon: I agree

sebbacon:
For testing, I wonder if @wjchulme 's [code counts notebook](https://github.com/opensafely/primary-care-covid-codes-research) might be good.

At the moment the study definition takes the earliest date of each event, per-patient, which is slightly different from what would be possible with the measures version, so maybe not... I'll discuss with him

sebbacon:
Documentation - I guess this should just go in the documentation repo pending decisions about docstrings? Or is there an obvious place to maintain it in the code?

sebbacon (approved):

wjchulme:
Some notes on using this for real:

#### index-date-range
The argument for `index-date-range` must be given inside `"`, not `'`. Eg, using 

     cohortextractor generate_cohort --expectations-population 10000 --index-date-range '2020-02-01 to 2020-09-01 by month'

fails with `unrecognized arguments: to 2020-09-01 by month'`. doing 

    cohortextractor generate_cohort --expectations-population 10000 --index-date-range "2020-02-01 to 2020-09-01 by month"

instead will work. This doesn't matter except the documentation notes above need updating.

#### mixed date ranges

If you run

    cohortextractor generate_cohort --expectations-population 10000 --index-date-range "2020-02-01 to 2020-09-01 by month"

and 

    cohortextractor generate_cohort --expectations-population 10000 --index-date-range "2020-02-01 to 2020-09-01 by week"


then 1)  `input_2020-02-01.csv` from the first month run will be overwritten by the second week run. 2) `generate_measures` doesn't know the difference between the outputs from these two runs, and will combine everything. 

Also, if there is an overall `/output/input.csv` file as well as the `/ouput/input_2020-02-01.csv` etc files, then running

    cohortextractor generate_measures

will fail with `'NoneType' object has no attribute 'group'` (since `input.csv` doesn't have a suffix). 

All of these issues can be dealt with if the `input_<date>.csv` files are put in a subfolder, perhaps taking the name `ouput/2020-02-01 to 2020-09-01 by week` for clarity.

PR #280, opened 2020-09-01
ACME is EMIS

inglesp:
This should make `emis_backend` capable of supporting everything that the factors-associated-with study requires.

@evansd let me know if you'd like some background on the `hashed_organisation` stuff.


cohortextractor/emis_backend.py
```diff
@@ -744,73 +744,51 @@ def _none_of_these_codes_occur_on_same_day(self, joined_table, codelist):
...
 
     def patients_address_as_of(self, date, returning=None, round_to_nearest=None):
-        # N.B. A value of -1 indicates no postcode recorded on the
-        # record, an invalid postcode, or no fixed abode.
-        #
-        # Related, there is a column in the address table to indicate
-        # NP for no postcode or NFA for no fixed abode
+        # At the moment we can only return current values for the fields in question.
+
+        if date != "today":
```
evansd, line 798: I was surprised this worked at all because since https://github.com/opensafely/cohort-extractor/pull/255/commits/bd9c7be9ab440a97d86803e7fc4d87805bb9bef9 date expressions are all supposed to be transformed into ISO strings by the time they reach the backend. What I've realised is that this isn't working for `address_as_of` because it uses the argument `date` rather than `reference_date` which is a bug that I need to fix. But when that's fixed this will stop working.

The simplest change to retain roughly the same behaviour would be to calll `datetime.date.today()` here and compare the result. Maybe we could relax the condition slightly and just require that the date is within a certain distance (one month?) of today.

cohortextractor/emis_backend.py
```diff
@@ -744,73 +744,51 @@ def _none_of_these_codes_occur_on_same_day(self, joined_table, codelist):
         """
 
     def patients_registered_practice_as_of(self, date, returning=None):
+        # At the moment we can only return current values for the fields in question.
```
evansd, line 772: Does this need a similar check on the date as below?

evansd (approved):
This all looks great. Very easy to review commit-by-commit. (The organisation hash stuff looks like it was pretty painful, but your explanation and the associated changes all made sense.)

There's a bit of faff with dates which I've commented on below, but you could merge this as is and then I could make the relevant changes at the point where I fix the bug in my code.

cohortextractor/emis_backend.py
inglesp: Thanks.

cohortextractor/emis_backend.py
inglesp: Good idea, I've fixed this -- could you re-review when you get the chance?

cohortextractor/emis_backend.py
```diff
@@ -999,6 +995,13 @@ def get_db_connection(self):
         self._db_connection = presto_connection_from_url(self.database_url)
         return self._db_connection
 
+    def validate_recent_date(self, date, max_delta_days=30):
+        date = datetime.datetime.strptime(date, "%Y-%m-%d").date()
```
evansd, line 999: You can do `datetime.date.fromisoformat()` these days

evansd (approved):
Looks great!

PR #284, opened 2020-09-02
Include non-ventilated patients in ICU results

### LINKED ISSUE #275: ICNARC function should not exclude non-ventilated patients
See discussion: https://github.com/opensafely/cohort-extractor/issues/273#issuecomment-679052963

Users need to be able to distinguish between ventilated and non-ventilated subgroups within ICNARC data.

We also need to consider which previous/current studies use ICNARC as an outcome and decide if they need to be re-run.

@HelenCEBM I think when I first made ICNARC I copied some code from a notebook you wrote without really thinking about why. Can you remember what the reason was for excluding patients without respiratory support in the first place? Just wondering if there was a reason to believe they were bad data.

evansd:
We have decided to ignore the value of the `Ventilator` column as we don't trust it. See discussion at:
https://github.com/opensafely/cohort-extractor/issues/275#issuecomment-685684964

Instead we use the counts for days of basic and advanced respiratory support (exact meaning of these tbc with ICNARC).

Returning options are now: `had_basic_respiratory_support`, `has_advanced_respiratory_support` and `had_respiratory_support` (which covers either case).

Closes #275

BREAKING CHANGE:
 1. Previously non-ventilated patients were excluded entirely from ICU results
 2. The `was_ventilated` returning option has been removed.

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -279,8 +279,12 @@ def admitted_to_icu(
     Options for `returning` are:
 
         binary_flag: Whether patient attended A&E
-        was_ventilated: Whether patient was ventilated
         date_admitted: Date patient arrived in A&E
+        had_respiratory_support: Whether patient received any form of respiratory support
+        had_basic_respiratory_support": Whether patient received "basic" respiratory support
+        had_advanced_respiratory_support": Whether patient received "advanced" respiratory support
+
+        (Note that the terms "basic" and "advanced" are derived from the underlying ICNARC data.)
     """
 
     return "admitted_to_icu", locals()
--- a/cohortextractor/process_covariate_definitions.py
+++ b/cohortextractor/process_covariate_definitions.py
@@ -340,7 +340,9 @@ class GetColumnType:
             "household_size": "int",
             "date_arrived": "date",
             "discharge_destination": "str",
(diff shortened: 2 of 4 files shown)


evansd:
It occurs to me that my semver "breaking change" may push us from version 0 to version 1. Personally I don't care about this but I'm aware some people attach signalling value to version 1.0 and so I thought I should flag this up.

sebbacon:
Looks good. Can you quickly search existing repos to check if any of them need changing to reflect this (and notify the author)? (I don't think they do).

> my semver "breaking change" may push us from version 0 to version 1

I think our decision to embrace semver inevitablly means we can no longer attach any romantic feelings to particular numbering systems.

evansd:
Have messaged the channel and will await some response before merging.

evansd:
Following discussion I'm going to do further work on this before merging. I think that now we've decided not to use the `Ventilator` column (because we don't trust it) we need to rename the `was_ventilated` returning value. I think we now need something like `had_basic_respiratory_support`, `had_advanced_respiratory_support`, and `had_respiratory_support` (to cover either case). Otherwise there's all sorts of potential for confusion.

sebbacon (approved):

PR #290, opened 2020-09-08
Fix transaction handling

evansd:
pyODBC and cTDS have quite different ideas about transaction management:
https://github.com/mkleehammer/pyodbc/wiki/Database-Transaction-Management
https://zillow.github.io/ctds/transactions.html

I had managed not to notice the bug here because the "happy path" for using temporary tables (where the download works first time and the temporary table can be immediately deleted) wasn't previously tested.


sebbacon (approved):
Different definitions of "auto" commit is not something you particularly plan for..!

PR #292, opened 2020-09-10
Fetch results in batches to limit memory usage

evansd:
Because the server this usually runs on has so much memory this would
work anyway in some cases so we didn't immediately notice. But it would
intermittently fail with `MemoryError` and certainly needs fixing.


cohortextractor/mssql_utils.py
```diff
@@ -88,3 +89,37 @@ def mssql_sqlalchemy_engine_from_url(url):
...
+    iterator = _dbapi_cursor_to_csv_file(cursor, filename, batch_size=batch_size)
+    # We don't want the default behaviour of this function to be to create an
+    # iterator which does nothing until you consume it (which might lead to
+    # confusion) but we do want the option of yielding the results so that
+    # callers who need to run additional checks on the output can do so
+    if yield_output_rows:
+        return iterator
+    else:
+        for line in iterator:
+            pass
```
inglesp, line 110: Given that we don't ever call `dbapi_cursor_to_csv_file` with `yield_output_rows=False`, do we need the `yield_output_rows` argument?  It's bit head-scratchy.

cohortextractor/tpp_backend.py
```diff
@@ -42,14 +42,16 @@ def to_csv(self, filename):
...
+        result_cursor = self.execute_queries(queries)
+        # `batch_size` here was chosen through a bit of unscientific
+        # trial-and-error and some guesswork. It may well need changing in
+        # future.
+        rows = dbapi_cursor_to_csv_file(
+            result_cursor, temp_filename, batch_size=32000, yield_output_rows=True
+        )
+        for row in rows:
+            # The first column contains IDs
+            unique_check.add(row[0])
```
inglesp, line 54: Would it be conceptually simpler to produce the CSV files first, and then do the uniqueness check?

inglesp (approved):

cohortextractor/mssql_utils.py
evansd: Ah yeah, I just copied it back from the old `mssql_query_to_csv_file` function which used the same pattern. But you're right, there's not really any need for it.

cohortextractor/mssql_utils.py
```diff
@@ -88,3 +89,25 @@ def mssql_sqlalchemy_engine_from_url(url):
...
+    with open(filename, "w", newline="") as csvfile:
+        writer = csv.writer(csvfile)
+        writer.writerow([x[0] for x in cursor.description])
+        while True:
+            result_batch = cursor.fetchmany(batch_size)
+            if not result_batch:
+                break
+            for row in result_batch:
+                row_callback(row)
+                writer.writerow(row)
```
inglesp, line 113: Do we want these two lines the other way round?  If the callback fails, it'd be nice to know which line it failed on.

inglesp (approved):

cohortextractor/mssql_utils.py
evansd: Ah, good thinking!

PR #300, opened 2020-09-22
Fixes to measure framework

### LINKED ISSUE #299: There is already an object named '#codelist_1_previous_stroke' in the database
Error running `generate_cohort --index-date-range "2020-02-01 to today by month"` on the server. 

    _tdsProgrammingError: There is already an object named '#codelist_1_previous_stroke' in the database

It didn't fail on dummy data locally. It doesn't fail when only running `generate_cohort`.

The study definition is here: https://github.com/opensafely/vacc-trial-monitoring-research/blob/measures-server/analysis/study_definition.py 

DSC_0207

evansd:
* **Reinitialise backend when changing index date**
   This is obviously necessary but didn't get spotted before because the
   tests use the dummy data generator.

 * **Ignore files without dates in the names**
   Previously, if there was an `input.csv` file present it would cause the
   `generate_measures` command to throw an error.

 * **Support not setting `group_by`**
    This is equivalent to "group everything"

Fixes #299 and addresses some of the issues raised in https://github.com/opensafely/cohort-extractor/pull/269#issuecomment-696046829

### DIFF
--- a/analysis/study_definition.py
+++ b/analysis/study_definition.py
@@ -156,6 +156,11 @@ study = StudyDefinition(
 
 measures = [
     Measure(
+        id="liver_disease",
+        numerator="has_chronic_liver_disease",
+        denominator="population",
+    ),
+    Measure(
         id="liver_disease_by_stp",
         numerator="has_chronic_liver_disease",
         denominator="population",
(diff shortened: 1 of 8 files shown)


sebbacon (approved):
Grand!

Do we need to update documentation? I presume the only place is the documentation repo - there's no obvious place in the code...

evansd:
Yes, changes need to go in the docs repo. I've got a separate PR for that.

PR #304, opened 2020-09-29
Improve performance of "not on same day" queries

### LINKED ISSUE #302: Slow-running cohort extraction
**Problem:**
- Slow-running cohort extractor for flu vaccine dashboards when hitting [this SQL step](https://gist.github.com/sebbacon/b652e77e7762777dd74bf038281e5ed7). 
- It has previously run in 1-2 hours but last week after several attempts of multiple hours it eventually failed overnight with `general SQL server error: check messages from the SQL server`. 

**What is going on:**
- TPP found that the SQL has run on different occasions with "vastly different execution plans" and is "at the mercy of the query optimiser".

**Solutions**:
- [ ] Bulid time-stamps into traceback to better identify how long queries ran before failing.
- [ ] Implement the following SQL into the back end, written by Jonny, which runs in a few minutes:
```
SELECT
   Patient_ID AS patient_id,
   1 AS has_event,
   MIN(ConsultationDate) AS ConsultationDate
INTO
#vaccs_codelist_2
FROM
CodedEvent
INNER JOIN #codelist_2_flu_vaccine_clinical_current
ON
(
CTV3Code = #codelist_2_flu_vaccine_clinical_current.code
)
WHERE
ConsultationDate BETWEEN '20200901' AND '20201031'
GROUP BY
Patient_ID
--(105445 row(s) affected)

CREATE CLUSTERED INDEX ix ON  #vaccs_codelist_2 (Patient_ID, ConsultationDate)

SELECT
Patient_ID,
CAST(ConsultationDate AS date) AS ConsultationDate
INTO
#vaccs_codelist_3
FROM
CodedEvent
INNER JOIN #codelist_3_flu_vaccine_clinical_current
ON
(
CTV3Code = #codelist_3_flu_vaccine_clinical_current.code
)
--(6921339 row(s) affected)

CREATE CLUSTERED INDEX ix ON  #vaccs_codelist_3 (Patient_ID, ConsultationDate)

-- Query for flu_vaccine_clinical_current
SELECT * INTO #flu_vaccine_clinical_current
FROM
(
SELECT *
FROM
#vaccs_codelist_2 v2
WHERE
NOT EXISTS
(
SELECT 1
FROM
#vaccs_codelist_3 v3
WHERE
v3.Patient_ID = v2.Patient_ID
AND v2.ConsultationDate = v3.ConsultationDate
)
) t
--(105378 row(s) affected)
```

evansd:
This PR improves the performance of queries which use the `ignore_days_where_these_codes_occur` option. It does this by creating a temporary table with an index on it, and then using that in the subquery rather than the enormous `CodedEvent` table.

This requires some refactoring to support creating arbitrary temporary tables. The handling of codelist tables could be subsumed into this which would simplify things but we leave that refactor for another day.

Closes #302

### DIFF
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -735,7 +735,7 @@ class TPPBackend:
         codelist_table = self.create_codelist_table(codelist, codes_are_case_sensitive)
         date_condition = make_date_filter("ConsultationDate", between)
         ignored_day_condition, extra_queries = self._these_codes_occur_on_same_day(
-            from_table, ignore_days_where_these_codes_occur
+            from_table, ignore_days_where_these_codes_occur, date_condition
         )
 
         # Result ordering
@@ -824,7 +824,7 @@ class TPPBackend:
         codelist_table = self.create_codelist_table(codelist, case_sensitive=False)
         date_condition = make_date_filter("ConsultationDate", between)
         ignored_day_condition, extra_queries = self._these_codes_occur_on_same_day(
-            "MedicationIssue", ignore_days_where_these_codes_occur
+            "MedicationIssue", ignore_days_where_these_codes_occur, date_condition
         )
         if episode_defined_as is not None:
             pattern = r"^series of events each <= (\d+) days apart$"
@@ -876,7 +876,7 @@ class TPPBackend:
         codelist_table = self.create_codelist_table(codelist, case_sensitive=True)
         date_condition = make_date_filter("ConsultationDate", between)
         ignored_day_condition, extra_queries = self._these_codes_occur_on_same_day(
-            "CodedEvent", ignore_days_where_these_codes_occur
+            "CodedEvent", ignore_days_where_these_codes_occur, date_condition
         )
         if episode_defined_as is not None:
             pattern = r"^series of events each <= (\d+) days apart$"
@@ -915,7 +915,7 @@ class TPPBackend:
         """
         return ["patient_id", "episode_count"], sql, extra_queries
 
-    def _these_codes_occur_on_same_day(self, joined_table, codelist):
+    def _these_codes_occur_on_same_day(self, joined_table, codelist, date_condition):
         """
         Generates a SQL condition that filters rows in `joined_table` so that
         they only include events which happened on days where none of the codes
@@ -937,6 +937,7 @@ class TPPBackend:
             FROM CodedEvent
             INNER JOIN {codelist_table}
             ON CTV3Code = {codelist_table}.code
+            WHERE {date_condition}
             """,
             f"""
             CREATE CLUSTERED INDEX ix ON {same_day_table} (Patient_ID, day)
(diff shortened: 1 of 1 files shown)


cohortextractor/tpp_backend.py
```diff
@@ -234,8 +235,9 @@ def get_queries(self, covariate_definitions):
                 )
             else:
                 date_format_args = pop_keys_from_dict(query_args, ["date_format"])
-                cols, sql = self.get_query(name, query_type, query_args)
+                cols, sql, setup_sql = self.get_query(name, query_type, query_args)
                 table_queries[name] = f"SELECT * INTO #{name} FROM ({sql}) t"
+                table_setup_queries[name] = setup_sql
```
inglesp, line 241: Looks like `setup_sql` will  be iterated over -- can that be reflected in its name?

inglesp (approved):
All makes sense.  It makes the overall flow a bit harder to follow, but I think that's necessary.  And it means it diverges yet further from the EMIS backend, but there's nothing we can do about that.

cohortextractor/tpp_backend.py
```diff
@@ -316,6 +324,12 @@ def get_query(self, column_name, query_type, query_args):
         self._current_column_name = column_name
         return_value = method(**query_args)
         self._current_column_name = None
+        # We want to support returning a (columns, sql, setup_queries) triple
+        # from query methods but without, for now, having to rewrite all the
+        # old methods which just return (columns, sql). So we catch old-style
+        # return values here and add an empty `setup_queries` list to the end.
+        if isinstance(return_value, tuple) and len(return_value) == 2:
```
sebbacon, line 332: I'm thinking about the refactoring we know we want to do, but not just yet, and how we signal that for future us.  In particular it would be good for legacy callers to be marked as such, somehow. We could consider a deprecation warning approach, but that means we'd also want to incorporate some kind of checks to surface them; or we could add comments in all the callers; or make a ticket (problem: we'll probably forget it exists). Any thoughts?

cohortextractor/tpp_backend.py
```diff
@@ -909,13 +909,13 @@ def _number_of_episodes_by_clinical_event(
...
         ) t
         GROUP BY Patient_ID
         """
         return ["patient_id", "episode_count"], sql
 
-    def _none_of_these_codes_occur_on_same_day(self, joined_table, codelist):
+    def _these_codes_occur_on_same_day(self, joined_table, codelist):
         """
         Generates a SQL condition that filters rows in `joined_table` so that
         they only include events which happened on days where none of the codes
```
sebbacon, line 921: This docstring is not not incorrect

sebbacon (approved):

cohortextractor/tpp_backend.py
evansd: In this case, it's not that there's a new shiny function signature and a deprecated one and we want to move everything over to the new one. Rather, both are a fudge at the moment and I think the right refactor involves a bigger change.

I'll write up what I think needs doing in a ticket. We might well forget about it, but it's better than nothing.

PR #309, opened 2020-10-02
Refactor codelist handling in TPP backend

### LINKED ISSUE #305: Refactor uploading of codelists in TPP backend
This would remove the added complexity introduced in #304 and generally make the code a bit easier to understand.

Specifically, what I think needs to happen is:
 * Covariate methods (the various `patient_` methods) should stop returning the column names of the query they generate and should instead produce queries with predictably named columns. In most cases this can be the value of `returning` e.g. `binary_flag` or `date_of_admission`. Where this isn't applicable we could use something generic like `value`.
 * Instead of returning an SQL string and then, optionally, a list of "setup queries" which need to be run first, these methods should return a single list of queries. The final member of this list is the SELECT query used to generate the output, and previous members (if any) are the setup queries that need to be run first.
 * The mechanism whereby codelists to be uploaded are tracked in a separate attribute on the class can be replaced by using the setup queries. i.e. if a particular covariate wants to create some codelists in a temporary tables it just returns the appropriate setup queries. This means that codelists no longer need any special handling.

Final note: we should make sure that setup queries have a comment as their first line explaining what they're doing. These will be picked up by the `execute_queries()` method and output as part of the log so the user gets some visibility into what the cohortextractor is doing.

evansd:
As well as having a net-negative line count this simplifies some of the internal data structures and generally (I think) gives the code a slightly clearer conceptual model.

Closes #305

### DIFF
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -98,8 +98,7 @@ class TPPBackend:
         if len(queries):
             # Add comment to first query
             queries[0] = f"-- Write codelists into temporary tables\n\n{queries[0]}"
-        for name, query in self.queries:
-            queries.append(f"-- Query for {name}\n{query}")
+        queries.extend(self.queries)
         return queries
 
     def save_results_to_temporary_db(self, queries):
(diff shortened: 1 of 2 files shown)


sebbacon (approved):
I love this so much!

Issue #297, opened 2020-09-14
BMI variable appears non-deterministic in some cases


evansd:
i.e. running the same study definition on the same data produces different outcomes.

This seems to be relatively rare: in a cohort of over 730,000 there were 5 affected results. And in all but one of those results the variation in value was small. In one case the value would sometimes drop to zero, indicating that no BMI was available.

Current theory is that this in an issue with this query here (and the equivalent height and BMI queries):
https://github.com/opensafely/cohort-extractor/blob/da036416bbb2f55bb4fbbca1b791acdf16f0f2d1/cohortextractor/tpp_backend.py#L483-L490

If a patient has several values recorded with exactly the same `ConsultationDate` then the ORDER clause is not sufficient to make the results deterministic.

-----

Note to self: there are files in `/e/dave-evans/T1DM_covid_research/outputs` which demonstrate the problem and provide specific examples for further investigation.

PR #315, opened 2020-10-09
Add ORDER BY clauses to make results deterministic

evansd:
In certain cases where we use the `ROWNUMBER() OVER (PARTITION BY ..`
`ORDER BY .. )` pattern to select only the first matching row in a join,
we don't specify enough order conditions to uniquely determine a first
row. When this happens the result is arbitrary (which is sort of OK) and
non-deterministic (which is not OK). This patch attempts to fix the
non-determinism by adding a tie-breaker column to the end of the sort
conditions, which is usually the ID field of the table in question.

There are two cases (the EC and APCS tables) where we can't 100%
guarantee determinstic ordering because this are no fields with a unique
constraint. However in both cases there is an "ident" field which is
almost entirely unique and where the tiny handful of duplicates are
otherwise identical and so shouldn't result in non-deterministic outputs
anyway.

This ought to fix the issue identified in #297 although I haven't yet
confirmed this.

Note that while this fixes the non-determinism it doesn't address the
arbitrariness. That is, where we have multiple matches there might be
better reasons for choosing one result over another, or better ways of
combining those results together as in the case of BMI.

### DIFF
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -993,7 +993,9 @@ class TPPBackend:
         # Note that current addresses are recorded with an EndDate of
         # 9999-12-31. Where address periods overlap we use the one with the
         # most recent start date. If there are several with the same start date
-        # we use the longest one (i.e. with the latest end date).
+        # we use the longest one (i.e. with the latest end date). We then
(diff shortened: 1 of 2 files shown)


inglesp (approved):
Looks good.

sebbacon (approved):

PR #316, opened 2020-10-09
fix: Remove Stata model tests from CI jobs

evansd:
This doesn't really belong here anymore as it doesn't test anything
meaningful about the cohortextractor. Also it's currently broken because
of problems downloading the Stata image (did we move it?).


inglesp (approved):

sebbacon:
I approve, but I think this means you could also remove `analysis/`?

sebbacon (approved):

evansd:
I think we use the study def as a smoke test fixture, but I can move it into `tests` now where it belongs.

PR #318, opened 2020-10-12
build: add ghcr.io publish action

bloodearnest:


.github/workflows/docker-publish-ghcr.yml
```diff
@@ -0,0 +1,77 @@
...
+      - uses: actions/checkout@v2
+
+      - name: Run tests
+        run: |
+          if [ -f docker-compose.test.yml ]; then
+            docker-compose --file docker-compose.test.yml build
+            docker-compose --file docker-compose.test.yml run sut
+          else
+            docker build . --file Dockerfile
+          fi
```
evansd, line 36: This seems a bit redundant. Both because this only runs on merge to master at which point all the tests should have run anyway, and also because the only thing it's testing is that `docker build` works which we have to run below anyway.

If there was someway of re-using the built image from the test in the push step then I'd be all for it, but as it is it seems like it would just double the length of time it takes to get a new image pushed without much benefit.

evansd (approved):

.github/workflows/docker-publish-ghcr.yml
bloodearnest: Yeah, I'm new to actions, so not sure how the best fit together. This was in the marketplace 'push to ghcr.io' action, so I left it in. I've no objection to removing the compose bit, but it seems that checking the image can build before merging is a good idea.

I'm mostly trying to get a basic deploy of a public image to ghcr.io, so I can check the auth interactions.

It's possible the build layers should be cached and thus fast for the rebuild for the push step. It may possible to share large output artifacts between jobs.  I will take a look and try clarify.

Assuming the test step runs against the branch, however, the tree built for testing may not exactly match the tree used to build post merge, at least commit wise, so we may have to re-build post merge.

If we drop the test, then can we still retry a broken post-merge triggered publish that failed?

PR #321, opened 2020-10-15
feat: Support returning `Household.TppPercentage` column

### LINKED ISSUE #313: % TPP coverage in each household
This is a request to add a variable which supplements the existing "MixedSoftwareHousehold" variable (https://github.com/opensafely/cohort-extractor/issues/285). The MixedSoftwareHousehold variable is created by TPP by comparing data in the MPI/UPRN tables.  This is described here: https://docs.google.com/document/d/1EzaRTiapjxxbj10wjN5iYjXbeyHMEErOoaV0tH6Mv1c/edit

The resulting binary variable is then read into OS as described in the closed issue 285 (https://github.com/opensafely/cohort-extractor/issues/285). 

The issue with only having a binary variable is that a large % of care homes are currently being marked as mixed software, but we have no idea if this is because there's a difference of several hundred people, or just one person. It'd be more informative to know the % coverage. 

Due to the potential disclosure risk of the raw counts, we propose to add another variable called something like "MixedSoftwareCoverage" which simply returns the proportion of individuals within a household that is estimated to be covered by TPP using this method. That would allow us to tell whether, for example, 90% of individuals in a care home are covered by TPP in which case we can likley use data from that care home. 

It is likely most efficient for this to be created by TPP and added to the household table, and then for this to be pulled in (similar to the way the MixedSoftwareHousehold variable was added).

evansd:
This is accessed via the slightly verbose option name:
percentage_of_members_with_ehr_data_available

This means it can be used across other EHR systems and is hopefully
self-explanatory to someone encountering it in a study definition for
the first time.

Example:
```
percent_tpp=patients.household_as_of(
    "2020-02-01", returning="percentage_of_members_with_ehr_data_available"
),
```

Closes #313

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -586,6 +586,12 @@ def household_as_of(reference_date, returning=None, return_expectations=None):
                                           different EHR system, meaning that our
                                           coverage of the household is incomplete.
 
+        percentage_of_members_with_ehr_data_available: Integer giving the (estimated)
+                                                       percentage of household members
+                                                       where we have EHR data available
+                                                       (i.e. not in other systems as
+                                                       above)
+
         msoa: Returns the MSOA (Middle Super Output Area) in which the household is
               situated
 
--- a/cohortextractor/process_covariate_definitions.py
+++ b/cohortextractor/process_covariate_definitions.py
@@ -341,6 +341,7 @@ class GetColumnType:
             "primary_diagnosis": "str",
             "is_prison": "bool",
             "has_members_in_other_ehr_systems": "bool",
+            "percentage_of_members_with_ehr_data_available": "int",
             "msoa": "str",
         }
         try:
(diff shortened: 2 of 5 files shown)


sebbacon (approved):
Pleasingly brief and satisfyingly generically named!

esnightingale:
I can't seem to set reasonable expectations for this variable - have tried normal distribution with type float but keep getting an error that "Column definition percent_tpp does not return expected type int"

esnightingale:
My study definition code is:

`    percent_tpp=patients.household_as_of(
        "2020-02-01",
        returning="percentage_of_members_with_ehr_data_available",
        return_expectations={"float": {"distribution": "normal", "mean": 75, "stddev": 10},
        },
    ),`

evansd:
Ah, confusingly it's type `int` rather than `float` to reflect the underlying TPP data.

esnightingale:
Will that not return an integer? I suppose that doesn't really matter

evansd:
It's the `return_expectations={"float"` bit that I think is causing the problem. I'm assuming `return_expectations={"int"` will just work. But if let me know if it doesn't and I'll take another look.

PR #322, opened 2020-10-19
Use batched downloads to work around "EOF" issue

evansd:
This PR uses a different approach to downloading results to try to work around the "EOF from server" error which has been occurring more frequently. We now download results in small batches which should (we think) be less likely to trigger the error and (more importantly) can be automatically retried if they fail so the user shouldn't have to do anything special to handle these errors.

### DIFF
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -31,33 +31,39 @@ class TPPBackend:
         self.queries = self.get_queries(self.covariate_definitions)
 
     def to_csv(self, filename):
-        queries = self.queries
+        queries = list(self.queries)
         # If we have a temporary database available we write results to a table
         # there, download them, and then delete the table. This allows us to
         # resume in the case of a failed download without rerunning the whole
         # query
         if self.temporary_database:
-            queries, cleanup_queries = self.save_results_to_temporary_db(queries)
+            output_table = self.save_results_to_temporary_db(queries)
         else:
-            cleanup_queries = []
+            output_table = "#final_output"
+            queries[-1] = (
+                f"-- Writing results into {output_table}\n"
+                f"SELECT * INTO {output_table} FROM ({queries[-1]}) t"
+            )
+            self.execute_queries(queries)
         temp_filename = self._get_temp_filename(filename)
         unique_check = UniqueCheck()
 
         def record_patient_id(row):
             unique_check.add(row[0])
 
-        result_cursor = self.execute_queries(queries)
         # `batch_size` here was chosen through a bit of unscientific
         # trial-and-error and some guesswork. It may well need changing in
         # future.
         dbapi_cursor_to_csv_file(
-            result_cursor,
+            self.execute_queries([f"SELECT * FROM {output_table}"]),
             temp_filename,
             batch_size=32000,
             row_callback=record_patient_id,
         )
-        if cleanup_queries:
-            self.execute_queries(cleanup_queries)
+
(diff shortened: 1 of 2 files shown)


cohortextractor/tpp_backend.py
```diff
@@ -31,33 +31,46 @@ def __init__(self, database_url, covariate_definitions, temporary_database=None)
...
-            queries, cleanup_queries = self.save_results_to_temporary_db(queries)
+            output_table = self.save_results_to_temporary_db(queries)
         else:
-            cleanup_queries = []
+            output_table = "#final_output"
+            queries[-1] = (
+                f"-- Writing results into {output_table}\n"
+                f"SELECT * INTO {output_table} FROM ({queries[-1]}) t"
+            )
+            queries.append(f"CREATE INDEX ix_patient_id ON {output_table} (patient_id)")
```
sebbacon, line 47: I wondered what difference, if any, making this `UNIQUE INDEX` would make. The answer is, [I think](https://explainextended.com/2009/04/23/making-an-index-unique/) "exactly none" in this case. Thought it was an interesting little article though!

cohortextractor/tpp_backend.py
```diff
@@ -31,33 +31,46 @@ def __init__(self, database_url, covariate_definitions, temporary_database=None)
...
+            table=output_table,
+            key_column="patient_id",
             batch_size=32000,
-            row_callback=record_patient_id,
+            row_callback=record_patient_id_and_log,
+            retries=2,
+            sleep=0.5,
+        )
+
+        self.execute_queries(
```
sebbacon, line 71: Do we want something like a `DROP TABLE IF EXISTS` in a `finally` clause here?

cohortextractor/tpp_backend.py
```diff
@@ -31,33 +31,46 @@ def __init__(self, database_url, covariate_definitions, temporary_database=None)
...
-            result_cursor,
+        mssql_table_to_csv(
             temp_filename,
+            cursor=self.get_db_connection().cursor(),
+            table=output_table,
+            key_column="patient_id",
             batch_size=32000,
-            row_callback=record_patient_id,
+            row_callback=record_patient_id_and_log,
+            retries=2,
```
sebbacon, line 67: I wonder if it'd be handy to make this an environment variable so it's easy to play with in production? (For that matter, perhaps the batch size too?)

sebbacon (approved):
All good, just some idle thoughts to add

cohortextractor/tpp_backend.py
evansd: One difference it would make is that, in the case that we accidentally generate duplicate patient rows, we'd blow up when trying to create the index rather than when the unique check runs (which is after we've downloaded the results). There might be an argument for doing it that way, but I don't want to make that kind of change right now.

cohortextractor/tpp_backend.py
evansd: In fact, we don't. When we've created a persistent table in the temporary database we want that to stick around in the case of error so that we can re-use the results. When we've only created a session-local table (using the `#` prefix) this will be cleaned up automatically when the connection closes. (In this case the final `DROP TABLE` is actually unnecessary, but it doesn't hurt and it means less branching in the code.)

cohortextractor/tpp_backend.py
evansd: Ah yeah, good shout. I'll do that later though.

PR #323, opened 2020-10-20
fix: Rename percentage_of_members_with_ehr_data_available to percenta…

inglesp:
…ge_of_members_with_data_in_this_backend

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -586,11 +586,11 @@ def household_as_of(reference_date, returning=None, return_expectations=None):
                                           different EHR system, meaning that our
                                           coverage of the household is incomplete.
 
-        percentage_of_members_with_ehr_data_available: Integer giving the (estimated)
-                                                       percentage of household members
-                                                       where we have EHR data available
-                                                       (i.e. not in other systems as
-                                                       above)
+        percentage_of_members_with_data_in_this_backend: Integer giving the (estimated)
+                                                         percentage of household members
+                                                         where we have EHR data available
+                                                         in this backend (i.e. not in other
+                                                         systems as above)
 
         msoa: Returns the MSOA (Middle Super Output Area) in which the household is
               situated
--- a/cohortextractor/process_covariate_definitions.py
+++ b/cohortextractor/process_covariate_definitions.py
@@ -341,7 +341,7 @@ class GetColumnType:
             "primary_diagnosis": "str",
             "is_prison": "bool",
             "has_members_in_other_ehr_systems": "bool",
-            "percentage_of_members_with_ehr_data_available": "int",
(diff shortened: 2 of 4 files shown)


cohortextractor/patients.py
```diff
@@ -586,11 +586,11 @@ def household_as_of(reference_date, returning=None, return_expectations=None):
                                           different EHR system, meaning that our
                                           coverage of the household is incomplete.
 
-        percentage_of_members_with_ehr_data_available: Integer giving the (estimated)
-                                                       percentage of household members
-                                                       where we have EHR data available
-                                                       (i.e. not in other systems as
-                                                       above)
+        percentage_of_members_with_data_in_this_ehr_system: Integer giving the (estimated)
```
evansd, line 589: This should be `percentage_of_members_with_data_in_this_backend`

evansd (changes requested):

evansd (approved):

evansd:
The only place this is used is:
https://github.com/opensafely/carehomes-research/blob/fd98cfc3279b184609349db1c084a8c7c2c3cc07/analysis/study_definition.py#L133-L138

evansd:
Copying rationale for this change from Slack conversation for future reference:

> inglesp Today at 11:27
> wrt https://github.com/opensafely/cohort-extractor/pull/321, is `percentage_of_members_with_ehr_data_available` misleading?
> Specifically, once we also have EMIS data, might researchers think that it means the percentage of members with EHR data available in TPP + EMIS?
> 
> dave  2 hours ago
> I see what you mean. We want something like `available_in_current_study` but it's already getting very long winded
> 
> seb  2 hours ago
> `percentage_of_members_with_backend_data_available` would be explicit for people who know the definition of "backend"
> 
> seb  2 hours ago
> `percentage_of_members_with_data_in_this_backend`
> :+1:
> 
> inglesp  2 hours ago
> I prefer long-winded to ambiguous
> 
> inglesp  2 hours ago
> Would you accept a PR that changes it?
> 
> seb  2 hours ago
> It's inevitable people are going to have to engage with the concept of a backend at some point, so as long as we're happy with that as the terminology, I prefer my latter suggestion
> 
> inglesp  2 hours ago
> I'm happy with that.  Dave?
> 
> dave  2 hours ago
> Yes, me too.
> 
> dave  2 hours ago
> I think there's only one study using this
> 
> seb  2 hours ago
> Thanks inglesp n
> 
> inglesp  44 minutes ago
> Should we rename has_members_in_other_ehr_systems to has_members_in_other_backends?
> 
> dave  40 minutes ago
> I don't think this suffers from quite the same ambiguity as the other example, and I feel like the proposed version is somehow less clear than the original (although I agree it's more consistent)
> 
> inglesp  38 minutes ago
> Fine

PR #324, opened 2020-10-20
Trial of google-style docstrings

sebbacon:
Following some iteration, these should eventually make it to
our documentation, automatically.

The main point of discussion is if there's benefit to moving to a more typed API sooner rather than later. For example, enums for `date_format`, custom types for `return_expectations`, etc.

On the one hand, this will tighten up our contracts and improve the documentation.

On the other hand, tightening up our contracts will force a major rationalisation / refactor which we know is coming but maybe isn't a top priority.

Here's how it renders in `mkdocstring`:

image

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -1156,6 +1156,53 @@ def with_tpp_vaccination_record(
     find_last_match_in_period=None,
     return_expectations=None,
 ):
+    """
+    Identify patients with a vaccination record for a target disease within the tpp vaccination record
+
+    Vaccinations can be recorded via a Vaccination Record or via prescription of a vaccine i.e a product code.
+
+    Args:
+        target_disease_matches: the target disease as a string
+        product_name_matches: the product name as a string
+        on_or_before: date of interest as a string with the format `YYYY-MM-DD`. Filters results to measurements
+            on or before the given date.The default value is `None`.
+        on_or_after: date of interest as a string with the format `YYYY-MM-DD`. Filters results to measurements
+            on or after the given date.The default value is `None`.
+        between: two dates of interest as a list with each date as a string with the format `YYYY-MM-DD`.
+            Filters results to measurements between the two dates provided. The default value is `None`.
+        returning: a string indicating what type of value should be returned. The options are limited to binary_flag
+            (which indicates if they have had the vaccination or not) or a date of vaccination
+        date_format: a string detailing the format of the dates to be returned. It can be "YYYY-MM-DD",
+            "YYYY-MM" or "YYYY" and wherever possible the least disclosive data should be returned. i.e returning
+            only year is less disclosive than a date with day, month and year.
+        find_first_match_in_period: a boolean that indicates if the data returned is first indication of vaccination
+            if there are multiple matches within the time period
+        find_last_match_in_period: a boolean that indicates if the data returned is last indication of vaccination
+            if there are multiple matches within the time period
+        return_expectations: a dictionary defining the incidence and distribution of expected value
+            within the population in question.
+
+    Returns:
+        list: of integers of `1` or `0` if `returning` argument is set to `binary_flag`;
+            list of strings with a date format returned if `returning` argument is set to `date`
+
+    Example:
+
+        A variable called `flu_vaccine` is created that returns the date of vaccination for
+        any patients in the GP dataset between 2 dates.
+
+            flu_vaccine=patients.with_tpp_vaccination_record(
+                target_disease_matches="influenza",
+                between=["2019-09-01", "2020-04-01"],
+                returning="date",
+                date_format="YYYY-MM",
+                find_first_match_in_period=True,
+                return_expectations={
+                    date": {"earliest": "2019-09-01", "latest": "2020-03-29"}
+                    }
+            ),
+    """
+
     return "with_tpp_vaccination_record", locals()
 
 
@@ -1177,6 +1224,46 @@ def with_gp_consultations(
     call. The concept of a "consultation" in EHR systems is generally broader
     and might include things like updating a phone number with the
     receptionist.
+
+    Args:
+        on_or_before: date of interest as a string with the format `YYYY-MM-DD`. Filters results to measurements
+            on or before the given date.The default value is `None`.
+        on_or_after: date of interest as a string with the format `YYYY-MM-DD`. Filters results to measurements
+            on or after the given date.The default value is `None`.
+        between: two dates of interest as a list with each date as a string with the format `YYYY-MM-DD`.
+            Filters results to measurements between the two dates provided. The default value is `None`.
+        find_first_match_in_period: a boolean that indicates if the data returned is first event
+            if there are multiple matches within the time period
+        find_last_match_in_period: a boolean that indicates if the data returned is last event
+            if there are multiple matches within the time period
+        returning: a string indicating what type of value should be returned. The options are limited to binary_flag
+            (which indicates if they have had the an event or not), date (which indicate date of event and used
+            with either find_first_match_in_period or find_last_match_in_period), or number_of_matches_in_period
+            (which counts the events in the period)
+        date_format: a string detailing the format of the dates to be returned. It can be "YYYY-MM-DD",
+            "YYYY-MM" or "YYYY" and wherever possible the least disclosive data should be returned. i.e returning
+            only year is less disclosive than a date with day, month and year.
+        return_expectations: a dictionary defining the incidence and distribution of expected value
+            within the population in question.
+
+    Returns:
+        list: of integers of `1` or `0` if `returning` argument is set to `binary_flag`;
+            list of strings with a date format returned if `returning` argument is set to `date`; a list of
+            integers if `returning` argument is set to number_of_matches_in_period
+
+    Example:
+
+        A variable called `gp_count` is created that counts number of GP consultation between two dates in
+        2019.
+
+            gp_count=patients.with_gp_consultations(
+                between=["2019-01-01", "2020-12-31"],
+                returning="number_of_matches_in_period",
+                return_expectations={
+                    "int": {"distribution": "normal", "mean": 6, "stddev": 3},
+                    "incidence": 0.6,
+                }
+            )
     """
     return "with_gp_consultations", locals()
 
@@ -1188,12 +1275,21 @@ def with_complete_gp_consultation_history_between(
     return_expectations=None,
 ):
     """
-    Because the concept of a "consultation" in EHR systems does not map exactly
-    to the GP-patient interaction we're interested in (see above) there is some
+    The concept of a "consultation" in EHR systems does not map exactly
+    to the GP-patient interaction we're interested in (see above) so there is some
     processing required on the part of the EHR vendor to produce the
     consultation record we need. This does not happen automatically as part of
     the GP2GP transfer, and therefore this query can be used to find just those
-    patients for which the full history is available.
+    patients for which the full history is available. This means finding patients
+    who have been continuously registered with a single TPP-using practice
+    throughout a time period.
+
+    Args:
+        start_date: start date of interest as a string with the format `YYYY-MM-DD`
+        end_date: end date of interest as a string with the format `YYYY-MM-DD`
+        return_expectations: a dictionary defining the incidence and distribution of expected value
+            within the population in question.
+
     """
     return (
         "with_complete_gp_consultation_history_between",
@@ -1218,13 +1314,9 @@ def with_test_result_in_sgss(
 ):
     """
     Finds lab test results recorded in SGSS (Second Generation Surveillance
-    System). Only SARS-CoV-2 results are included in our data extract so this
-    will throw an error if the specified pathogen is anything other than
-    "SARS-CoV-2".
+    System).
 
-    `test_result` must be one of: "positive", "negative" or "any"
-
-    The date field used is the date the specimen was taken, rather than the
+    Please note for the dates this is used in the database as the date the specimen was taken, rather than the
     date of the lab result.
 
     There's an important caveat here: where a patient has multiple positive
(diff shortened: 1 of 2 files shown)


inglesp (commented):
I'm +0.

It's good to have a consistent way of doing things.

Things I'd like to change:

* remove the long list of parameters in a large font after the function name -- everything's duplicated below, and it's overwhelming
* remove the type column -- it's odd to have eg a `str` which is the latest date, or a `list` which is a tuple (but not actually a tuple)
* have an example column instead
* make `returning` the first parameter

Also, there'll be some overlap between all of the function definitions, eg `on_or_before`.  Does it make sense to specify these in one place?  Or maybe that can come later.

sebbacon:
To clarify the main use-cases for doing this at all are:

* Versioned study definition documentation
* Making sure documentation updates with code changes

Thanks for your comments Peter, I think in the absence of custom typing you're right that we should just omit the types and lean mostly on the examples.

sebbacon:
Thanks Caroline. 

* Per Peter's suggestions I think we should drop the type information from the docstrings and rely just on the examples.  That would mean, for example, changing `with_these_diagnoses (codelist): icd10 codes to match against any diagnosis` to `with_these_diagnoses: a code list of icd10 codes to match against any diagnosis` (for example) - this needs doing in my commit too

General points (not necessarily specific to these commits, just thought worth mentioniing):

* As part of this effort, we should move any stray docstring from the backend (i.e. [here](https://github.com/opensafely/cohort-extractor/blob/master/cohortextractor/tpp_backend.py) to `patients.py`, *except* where something is backend-specific, in which case it should stay there, with an explicit note that it is such
* @evansd has done a great job of documenting new functions in PR comments when he adds them; worth going back through closed PRs to look for good examples etc 
* For reference, full Google-style docstring format is specified [here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) (and linked from there)

CarolineMorton:
> Thanks Caroline.
> 
> * Per Peter's suggestions I think we should drop the type information from the docstrings and rely just on the examples.  That would mean, for example, changing `with_these_diagnoses (codelist): icd10 codes to match against any diagnosis` to `with_these_diagnoses: a code list of icd10 codes to match against any diagnosis` (for example) - this needs doing in my commit too
> 
> General points (not necessarily specific to these commits, just thought worth mentioniing):
> 
> * As part of this effort, we should move any stray docstring from the backend (i.e. [here](https://github.com/opensafely/cohort-extractor/blob/master/cohortextractor/tpp_backend.py) to `patients.py`, _except_ where something is backend-specific, in which case it should stay there, with an explicit note that it is such
> * @evansd has done a great job of documenting new functions in PR comments when he adds them; worth going back through closed PRs to look for good examples etc
> * For reference, full Google-style docstring format is specified [here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) (and linked from there)

Ok thanks. Had pushed before I read this. Will have to pick this up when back next week. Would be interested to understand why we don't want to use typing. I think it makes it a bit easier to read but that is a personal preference. The examples are much more helpful.

CarolineMorton:
Final comment for tonight: What do you want to do re deprecated args? They are still in use, but I don't want to encourage them to be used particularly when these docstrings are pulled into the documentation? @evansd may have thoughts regarding the future of these args anyway?

madwort:
>     * remove the type column -- it's odd to have eg a `str` which is the latest date, or a `list` which is a tuple (but not actually a tuple)

Isn't that an argument for keeping the type column & getting around to improving the API? Because (e.g.) if you saw a column called `date` with no type information, one might assume it would want some kind of date type?

cohortextractor/patients.py
```diff
@@ -667,19 +878,37 @@ def admitted_to_hospital(
...
-    `with_these_procedures` is optional, and is a list of OPCS-4 codes
-
-    See https://github.com/opensafely/cohort-extractor/issues/186 for discussion.
+    """"
+    Return information about admission to hospital.
+
+    See https://github.com/opensafely/cohort-extractor/issues/186 for in-depth discussion and background.
+
+    Args:
+        on_or_before (str): The latest date of admission
```
madwort, line 887: what format is acceptable here? only `YYYY-MM-DD`?

cohortextractor/patients.py
```diff
@@ -667,19 +878,37 @@ def admitted_to_hospital(
...
+        on_or_after (str): The earliest date of admission
+        between (list): A tuple of [`on_or_after`, `on_or_before`]
```
madwort, line 889: you mean a list containing exactly one tuple? whuh?

cohortextractor/patients.py
```diff
@@ -667,19 +878,37 @@ def admitted_to_hospital(
...
+        returning (str): One of `binary_flag` (if they were admitted at all), `date_admitted`, `date_discharged`, `number_of_matches_in_period`, `primary_diagnosis`
+        find_first_match_in_period (bool): For date return values, always choose the earliest matching occurence
```
madwort, line 891: does this have a (meaningful) default thath should be noted? on the face of it, looks like default is `None` which is (iirc) falsey?

cohortextractor/patients.py
```diff
@@ -128,6 +295,50 @@ def mean_recorded_value(
...
+    us to account for multiple measurements taken on one day.
+
+    The date of the measurement can be included by flagging with date format options.
+
+    Args:
+        codelist (obj: Codelist): Codelist for requested value
+        on_most_recent_day_of_measurement (bool): flag for requesting measurements be on most recent date
+        return_expectations (dict, None): An expectation definition defining an incidence
+            and a distribution as a dict,
+        on_or_before (str, None): String of Data in YYYY-MM-DD format or None,
```
madwort, line 310: more type information, less meaning information

cohortextractor/patients.py
```diff
@@ -108,6 +241,40 @@ def most_recent_bmi(
     The date of the measurement can be obtained using `date_of("<bmi-column-name>")`.
     If the BMI is computed from weight and height then we use the date of the
     weight measurement for this.
+
+    Args:
+        on_or_before (str, None): String of Data in YYYY-MM-DD format or None,
```
madwort, line 246: (hasn't read the doc docs fully yet...) is there a way to abstract this out for reuse?

cohortextractor/patients.py
madwort: ah, seems like it's picking up the default automatically (this is the function from your example screenshot?)

cohortextractor/patients.py
```diff
@@ -128,6 +295,50 @@ def mean_recorded_value(
...
+    The date of the measurement can be included by flagging with date format options.
+
+    Args:
+        codelist (obj: Codelist): Codelist for requested value
+        on_most_recent_day_of_measurement (bool): flag for requesting measurements be on most recent date
+        return_expectations (dict, None): An expectation definition defining an incidence
+            and a distribution as a dict,
+        on_or_before (str, None): String of Data in YYYY-MM-DD format or None,
+        on_or_after (str, None): String of Data in YYYY-MM-DD format or None,
+        between (list, None): List of two date strings as YYYY-MM-DD format or None,
```
madwort, line 312: ...so if it's picking up the default automatically, what is `None` doing here?

sebbacon:
Maybe we should have a slack call to discuss the typing stuff. We could keep type info in, but choose not to show it in the docs.

On the other hand, for as long as our types are things like `list of dicts` or choices of strings, it's of limited value in many cases. For example we really want the different date format options to be some kind of `enum` -- this also overlaps with @madwort's question about refactoring for reuse.

Much of it boils down to this:

> Isn't that an argument for keeping the type column & getting around to improving the API? Because (e.g.) if you saw a column called date with no type information, one might assume it would want some kind of date type?

We do need to get onto rationalising the API but I don't want that to block us from writing minimally useful, versioned documentation that we can update as we do that.  That's the circle to square, really.

cohortextractor/patients.py
```diff
@@ -243,17 +243,33 @@ def most_recent_bmi(
...
+        include_measurement_date: This is a boolean flag (i.e. True or False) which if True allows us to
+            return an additional column of date of bmi. The default value is False.
+        date_format: This is a string detailing the format of the dates to be returned. It can be "YYYY-MM-DD",
+            "YYYY-MM" or "YYYY" and wherever possible the least disclosive data should be returned. i.e returning
+            only year is less disclosive than a date with day, month and year. The "include_measurement_date" flag
+            needs to be True for this to return a date.
+        include_month: This is a boolean flag for how granular to return date. In this case to return month as
+            well as year.. This is a deprecated argument, which is still supported for old versions and has been
+            replaced by the "date_format" argument.
+        include_day: This is a boolean flag for how granular to return date. In this case to return day as
```
sebbacon, line 270: Can we be consistent with how we construct the sentence, e.g. some start with "This is a" and some don't?

I think we should tend towards being concise i.e. dropping "this is a". For example ```include_day: a boolean indicating  if day should be included in addition to year and month (deprecated: use `date_format` instead)```

cohortextractor/patients.py
```diff
@@ -243,17 +243,33 @@ def most_recent_bmi(
...
-        between (list, None): List of two date strings as YYYY-MM-DD format or None,
-        minimum_age_at_measurement (int): Minimum age at measurement. Default value is 16.
-        return_expectations (dict, None): An expectation definition defining an incidence
-            and a distribution as a dict,
-        include_measurement_date (bool): Flag to return an additional column of date of bmi. Set as default to False
-        date_format (str): Date String of format to return date of measurement if flagged
-        include_month (bool): Flag for how granular to return date
-        include_day (bool): Flag for how granular to return date
-
+        on_or_before: Date of interest as a string with the format "YYYY-MM-DD". This will find data
```
sebbacon, line 246: What does "will find" mean? Can we be more specific? How about "restricts filter to consider data on or before the given date"?

cohortextractor/patients.py
```diff
@@ -243,17 +243,33 @@ def most_recent_bmi(
...
+            on or before the date of interest. The default value is None.
```
sebbacon, line 247: What does behaviour should the user expect if they supply `None` ?

cohortextractor/patients.py
```diff
@@ -243,17 +243,33 @@ def most_recent_bmi(
...
+        on_or_after: Date of interest as a string with the format "YYYY-MM-DD". This will find data
+            on or after the date of interest. The default value is None.
+        between: Two dates of interest as a list with each date as a string with the format "YYYY-MM-DD".
+            This will find data between the two dates only. The default value in None.
```
sebbacon, line 251: "*is* None"

cohortextractor/patients.py
sebbacon, line 247: Can we always quote literal values like `None` and `YYYY-MM-DD` and `1` with backticks?

cohortextractor/patients.py
```diff
@@ -243,17 +243,33 @@ def most_recent_bmi(
...
+        minimum_age_at_measurement: This is the minimum age at measurement of BMI. It is an integer
```
sebbacon, line 252: "This is the minimum age at measurement of BMI" doesn't expand enough on the keyword name. Something like "measurements taken before this age will not count towards BMI calculations"?

cohortextractor/patients.py
```diff
@@ -243,17 +243,33 @@ def most_recent_bmi(
...
+            and the default value is 16.
+        return_expectations: This is a dictionary defining the incidence and distribution of expected BMI
+            within the population in question. This is a 3-item key-value dictionary of "date" and "float".
+            "date" is dictionary itself and should contain the "earliest" and "latest" dates needed in the
+            dummy data. "float" is a dictionary of "distribution", "mean", and "stddev". These values determine
+            the shape of the dummy data returned, and the float means a float will be returned rather than an
+            integer. "incidence" must have a value and this is what percentage of dummy patients have
+            a BMI. It needs to be a number between 0 and 1.
+        include_measurement_date: This is a boolean flag (i.e. True or False) which if True allows us to
```
sebbacon, line 261: I don't think we need to explain what boolean means here. That belongs in non-tech documentation really.

sebbacon: "allows us to return an additional column of date of bmi" isn't explicit enough. What is the column called? "allows us" might imply there's another action required. Something like "a boolean indicating if an extra column, named `date_of_<colname>`, should be included in the output"?  (i've not checked that it's actually called `date_of_<colname>`!)

cohortextractor/patients.py
```diff
@@ -243,17 +243,33 @@ def most_recent_bmi(
...
+            return an additional column of date of bmi. The default value is False.
+        date_format: This is a string detailing the format of the dates to be returned. It can be "YYYY-MM-DD",
+            "YYYY-MM" or "YYYY" and wherever possible the least disclosive data should be returned. i.e returning
+            only year is less disclosive than a date with day, month and year. The "include_measurement_date" flag
+            needs to be True for this to return a date.
```
sebbacon, line 266: > The "include_measurement_date" flag needs to be True for this to return a date.

I think clearer to say "only used if `include_measurement_date` is `True`"? Otherwise it slightly implies it will return something other than a date if not `True`...

sebbacon (commented):
Bikeshedding as promised! Let me know if the rationale for any of my suggestions is unclear.

cohortextractor/patients.py
CarolineMorton: I am not sure "restricts filter to consider data on or before the given date" is that clear. What about "This  filters to data on or before the given date"

sebbacon: I see what you mean. "filters results to measurements on or before the given date"?

(I still have a preference for losing the word "this")

cohortextractor/patients.py
```diff
@@ -243,33 +243,31 @@ def most_recent_bmi(
...
             within the population in question. This is a 3-item key-value dictionary of "date" and "float".
             "date" is dictionary itself and should contain the "earliest" and "latest" dates needed in the
             dummy data. "float" is a dictionary of "distribution", "mean", and "stddev". These values determine
             the shape of the dummy data returned, and the float means a float will be returned rather than an
             integer. "incidence" must have a value and this is what percentage of dummy patients have
             a BMI. It needs to be a number between 0 and 1.
-        include_measurement_date: This is a boolean flag (i.e. True or False) which if True allows us to
-            return an additional column of date of bmi. The default value is False.
+        include_measurement_date: A boolean indicating if an extra column, named date_of_bmi,
+            should be included in the output. The default value is False.
```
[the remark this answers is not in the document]
sebbacon: `date_of_bmi` (backticks)

cohortextractor/patients.py
```diff
@@ -1,9 +1,8 @@
-"""
-These methods don't *do* anything; they just return their name and arguments.
-This provides a friendlier API then having to build some big nested data
-structure by hand and means we can make use of autocomplete, docstrings etc to
-make it a bit more discoverable.
-"""
+# These methods don't *do* anything; they just return their name and arguments.
+# This provides a friendlier API then having to build some big nested data
+# structure by hand and means we can make use of autocomplete, docstrings etc to
+# make it a bit more discoverable.
```
ghickman, line 4: Modules can have docstrings too!

cohortextractor/patients.py
```diff
@@ -1,9 +1,8 @@
-"""
-These methods don't *do* anything; they just return their name and arguments.
-This provides a friendlier API then having to build some big nested data
-structure by hand and means we can make use of autocomplete, docstrings etc to
-make it a bit more discoverable.
-"""
+# These methods don't *do* anything; they just return their name and arguments.
+# This provides a friendlier API then having to build some big nested data
```
ghickman, line 2: s/then/than/

ghickman (changes requested):

cohortextractor/patients.py
```diff
@@ -31,6 +59,29 @@ def age_as_of(
     # Required keyword
     return_expectations=None,
 ):
+    """
+    Returns age of patient of at a particular date
```
evansd, line 63: Note that age can be negative if a patient was born after the reference date!

evansd (approved):
This is EPIC! Amazing work! I know from experience how laborious this kind of documentation is, but also how important. I'm sure we'll end up refining and extending this in lots of ways further down the track, but this is definitely good to merge from my point of view.

evansd:
@CarolineMorton If you make one of your commit titles have the prefix `docs: ...` then it will pass the "Semantic Pull Request" check that's currently failing

cohortextractor/patients.py
CarolineMorton: thanks - i have corrected

cohortextractor/patients.py
CarolineMorton: fixed - thanks!

cohortextractor/patients.py
CarolineMorton: Added

CarolineMorton:
Thanks for the feedback! And for resolving my semantic PR issue!

cohortextractor/patients.py
```diff
@@ -1,27 +1,55 @@
...
-This provides a friendlier API then having to build some big nested data
-structure by hand and means we can make use of autocomplete, docstrings etc to
-make it a bit more discoverable.
-"""
-
+""" This module provides the methods for making a study definition """
+# These methods don't *do* anything; they just return their name and arguments.
+# This provides a friendlier API than having to build some big nested data
+# structure by hand and means we can make use of autocomplete, docstrings etc to
+# make it a bit more discoverable.
```
ghickman, line 5: Ah, sorry, I wasn't very clear last time!

This whole block can be a docstring.  The content is great and by making it a docstring means tooling can pick it up in the future 😄

ghickman (approved):

PR #331, opened 2020-11-02
CI: move the ghcr.io publish into the main action.

bloodearnest:
This serializes the publish to be able to use the auto-tagging steps so that we
can tag the docker image appropriately.


evansd (approved):

sebbacon:
Sorry for only half-paying attention, and pretty sure I've asked this before, but why couldn't we use `whoan/docker-build-with-cache-action` for building and publishing?

bloodearnest:
> Sorry for only half-paying attention, and pretty sure I've asked this before, but why couldn't we use `whoan/docker-build-with-cache-action` for building and publishing?

We possibly could, but a) wasn't sure it could publish to ghcr.io, and also wanted to try reduce dependency on external actions. 

That one in particular is complex.

sebbacon:
The one thing I liked about it is the promise of caching, but makes sense!

bloodearnest:
Right, I think caching of docker images in actions is the default now

PR #362, opened 2020-12-03
correct "find first and last match" argument docs

wjchulme:


CarolineMorton (approved):
Looks good. Thanks for fixing my boo-boo

PR #368, opened 2020-12-07
feat: Support multiple columns in measure `group_by`

### LINKED ISSUE #355: Group by multiple columns in 'Measure'
At the moment, the group_by option in Measure only supports use of a single variable, however, for many studies it'd be useful to be able to group by more than one variable. An example is in order to generate age and sex specific mortality rates, but it might also be valuable to be able to look at age-specific rates within different geographical areas, for example. 

I've discussed briefly on slack with @evansd , but let me know in case further information might be useful.

evansd:
Just supply a list or tuple of `group_by` columns e.g
```py
measures = [
    Measure(
        id="liver_disease_by_stp_and_sex",
        numerator="has_chronic_liver_disease",
        denominator="population",
        group_by=["stp", "sex"],
    ),
]
```

Closes #355

### DIFF
--- a/cohortextractor/cohortextractor.py
+++ b/cohortextractor/cohortextractor.py
@@ -288,7 +288,7 @@ def _generate_measures(output_dir, study_name, suffix, skip_existing=False):
 def _calculate_measure_df(patient_df, measure):
     if measure.group_by:
         measure_df = patient_df[
-            [measure.numerator, measure.denominator, measure.group_by]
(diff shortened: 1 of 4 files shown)


sebbacon (approved):
Question: is it worth adding docstring-based documentation here and pulling it in via mkdocs?

inglesp (approved):
Code looks good.

wjchulme (approved):

PR #369, opened 2020-12-08
Randomise dummy data patient_id

alexwalkercebm:
This way the `patient_id` is more representative of those in the real data, rather than a sequential list of integers. This fixes a specific issue with https://github.com/opensafely/matching, which drops the patients in one extract from another extract. Previously, the two extracts had identical `patient_id` in the dummy data, leading to all dummy patients being dropped, e.g. in [this error](https://github.com/opensafely/post-covid-thrombosis-research/runs/1524905879?check_suite_focus=true).

To consider: I can't remember if `patient_id` in the real data end up sorted, might be worth also sorting them here if so?

Also, re: the deleted comment, I'd argue that we always need the `patient_id` col in the output.

### DIFF
--- a/cohortextractor/study_definition.py
+++ b/cohortextractor/study_definition.py
@@ -3,6 +3,7 @@ import copy
 import os
 
 import pandas as pd
+from numpy.random import default_rng
 
 from .expectation_generators import generate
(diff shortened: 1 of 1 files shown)


evansd (approved):
Nice!

PR #370, opened 2020-12-10
Randomise dummy data patient_id

evansd:


evansd:
Originally opened so that tests would run (as they won't run on a PR from a fork) but tests seem to have run after I approved the PR.

PR #376, opened 2020-12-15
correct return values for sex

wjchulme:


CarolineMorton (approved):

inglesp:
@wjchulme to pass the Pedantic Pull Request check, you'll have to change the text in the commit itself.  If you use git on the command line, you can run `git commit --amend -m "fix: return values for sex"`.  You'll then have to force push, with `git push -f`.  Shout if none of this makes sense!

wjchulme:
@inglesp ty*10^10. I will inevitably fail to correctly name similar commits in future so bookmarking this PR

PR #378, opened 2020-12-16
Update EMIS backend for SRO work

inglesp:
This PR contains all the changes required to get the EMIS backend to produce SQL required for https://github.com/opensafely/restoration-observatory-emis-research.

### DIFF
--- a/cohortextractor/emis_backend.py
+++ b/cohortextractor/emis_backend.py
@@ -822,6 +822,40 @@ class EMISBackend:
             """,
         )
 
+    def patients_with_death_recorded_in_primary_care(
+        self,
+        # Set date limits
+        between=None,
+        # Set return type
+        returning="binary_flag",
+    ):
+        if returning == "binary_flag":
+            column = "1"
+        elif returning == "date_of_death":
+            column = "date_of_death"
+        else:
+            raise ValueError(f"Unsupported `returning` value: {returning}")
+        if between is None:
+            between = (None, None)
+        min_date, max_date = between
+        if max_date is None:
+            max_date = "9999-12-31"  # Far enough in the future to catch everyone
+        if min_date is None:
+            min_date = "1900-01-01"  # Far enough in the path to catch everyone
+        return (
+            ["patient_id", returning],
+            f"""
+        SELECT
+          registration_id AS patient_id,
+          hashed_organisation,
+          {column} AS {returning}
+        FROM
+          patient_view
+        WHERE
+          date_of_death BETWEEN {quote(min_date)} AND {quote(max_date)}
+        """,
+        )
+
     def patients_address_as_of(self, date, returning=None, round_to_nearest=None):
         # At the moment we can only return current values for the fields in question.
         self.validate_recent_date(date)
--- a/cohortextractor/emis_backend.py
+++ b/cohortextractor/emis_backend.py
@@ -15,9 +15,9 @@ safe_punctation = r"_.-"
 SAFE_CHARS_RE = re.compile(f"^[a-zA-Z0-9{re.escape(safe_punctation)}]+$")
 
 
-PATIENT_TABLE = "patient_view"
-MEDICATION_TABLE = "medication_view"
-OBSERVATION_TABLE = "observation_view"
+PATIENT_TABLE = "patient_all_orgs_v2"
+MEDICATION_TABLE = "medication_all_orgs_v2"
+OBSERVATION_TABLE = "observation_all_orgs_v2"
 ICNARC_TABLE = "icnarc_view"
 ONS_TABLE = "ons_view"
 CPNS_TABLE = "cpns_view"
--- a/cohortextractor/emis_backend.py
+++ b/cohortextractor/emis_backend.py
@@ -93,6 +93,11 @@ class EMISBackend:
                 output_columns[name] = self.get_column_expression(
                     column_type, **query_args
                 )
+            # As do `aggregate_of` columns
+            elif query_type == "aggregate_of":
+                output_columns[name] = self.get_aggregate_expression(
+                    column_type, output_columns, **query_args
+                )
             else:
                 date_format_args = pop_keys_from_dict(query_args, ["date_format"])
                 cols, sql = self.get_query(name, query_type, query_args)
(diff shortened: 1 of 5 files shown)


cohortextractor/emis_backend.py
```diff
@@ -1025,6 +1030,52 @@ def validate_recent_date(self, date, max_delta_days=30):
...
+                THEN {extreme_value}
+            ELSE {column_definitions[name]} END"""
+            for name in column_names
+        )
+
+        return f"""
+        CASE WHEN {function}({components}) = {extreme_value}
+            THEN {default_value}
+        ELSE {function}({components}) END"""
+
```
sebbacon, line 1132: So if I've understood this right, the single edge case that would be a bug is where there is one or more genuine values of `2 ** 63` that we want to find the `MIN` of, and vice-versa? 

Given this is exceedingly unlikely it seems the cost of adding a check exceeds the benefits, but I do wonder if a note should be added to the documentation somewhere, or a warning in the docstring?

cohortextractor/emis_backend.py
```diff
@@ -22,11 +22,17 @@ class EMISBackend:
     def __init__(self, database_url, covariate_definitions, temporary_database=None):
         self.database_url = database_url
         self.covariate_definitions = covariate_definitions
+        self.postprocess_covariate_definitions()
         self.codelist_tables = []
         self.temp_table_prefix = self.get_temp_table_prefix()
         self.log(f"temp_table_prefix: {self.temp_table_prefix}")
         self.queries = self.get_queries(self.covariate_definitions)
 
+    def postprocess_covariate_definitions(self):
```
sebbacon, line 39: docstring to explain?

cohortextractor/emis_backend.py
```diff
@@ -972,7 +972,11 @@ def patients_with_these_codes_on_death_certificate(
                 hashed_organisation,
                 {column_definition} AS {column_name}
             FROM ons_view
-            WHERE ({code_conditions}) AND {date_condition}
+            WHERE ({code_conditions})
+                AND {date_condition}
+                AND date_parse(upload_date, '%d/%m/%Y') = (
```
sebbacon: Add a comment here? 

Also if you've not already checked this, verify this is what TPP are already doing for us, and mention this behaviour in a short sentence [in the SGSS documentation](https://docs.opensafely.org/en/latest/dataset-sgsscovid/)?

sebbacon (approved):
This is bloody marvellous!

What happens to temporary tables after we've used them?

cohortextractor/emis_backend.py
inglesp: I don't think you have understood this right.  However, writing this all out has revealed a bug.

Firstly: we don't expect to handle having genuine values of `2 ** 63` at all.

Suppose we have the following pairs of values for three patients, and we want to find the minimum of each pair.

| id | value1 | value2 |
| - | - | - |
| 1 | 100 | 200 |
| 2 | 100 | null |
| 3 | null | null |
| 4 | 100 | 0 |

Here are the expected results:

| id | value1 | value2 | exp_min |
| - | - | - | - |
| 1 | 100 | 200 | 100 |
| 2 | 100 | null | 100 |
| 3 | null | null | 0 |
| 4 | 100 | 0 | 0 |

(We want a zero for row 3, because we replace nulls with a default value, and the default value for an integer column is zero.)

Earlier on in the process, we have already converted nulls to the default value, so when the SQL for this function is executed, we have:

| id | value1 | value2 | value1' | value2' | min(value1', value2') | exp_min
| - | - | - | - | - | - | -
| 1 | 100 | 200 | 100 | 200 | 100 | 100
| 2 | 100 | null | 100 | 0 | 0 | 100
| 3 | null | null | 0 | 0 | 0 | 0
| 4 | 100 | 0 | 100 | 0 | 0 | 0

Finding `min(value1', value2')` will give the wrong result for row 2.

So instead, we convert any default values to a large number:
| id | value1 | value2 | value1' | value2' | value1'' | value2'' | min(value1'', value2'') | exp_min
| - | - | - | - | - | - | - | - | -
| 1 | 100 | 200 | 100 | 200 | 100 | 200 | 100 | 100
| 2 | 100 | null | 100 | 0 | 100 | 9999 | 100 | 100
| 3 | null | null | 0 | 0 | 9999 | 9999 | 9999 | 0
| 4 | 100 | 0 | 100 | 0 | 100 | 9999 | 100 | 0

`min(value1'', value2'')` now gives the right result for row 2, but not for row 3 (which we fix) or row 4 (which we don't fix and which is the bug).

To fix for row 3, we can convert the large number back to the default value.

| id | value1 | value2 | value1' | value2' | value1'' | value2'' | m=min(value1'', value2'') | m == 9999 ? 0 : m | exp_min
| - | - | - | - | - | - | - | - | - | -
| 1 | 100 | 200 | 100 | 200 | 100 | 200 | 100 | 100 | 100
| 2 | 100 | null | 100 | 0 | 100 | 9999 | 100 | 100 | 100
| 3 | null | null | 0 | 0 | 9999 | 9999 | 9999 | 0 | 0
| 4 | 100 | 0 | 100 | 0 | 100 | 9999 | 100 | 100 | 0

This also affects the TPP implementation.  (I have a failing test case.)

inglesp: To summarise, if we're finding the minimum or maximum of a set of values, then any values that match the default value will be ignored, because we also use the default value instead of NULL.  This is most likely to affect finding the minimum values, because the default value is either the empty string (which is less than every other string), or zero (and most values, eg blood pressure, counts of things) are positive.

@evansd any bright ideas?

inglesp: [According to @evansd on Slack](https://ebmdatalab.slack.com/archives/C63UXGB8E/p1608211898093500), this is fine.  I have updated the comment.

cohortextractor/emis_backend.py
inglesp: Addressed in a7d1640

cohortextractor/emis_backend.py
inglesp: > verify this is what TPP are already doing for us

Done.

evansd (approved):
Very easy to review commit-by-commit, thanks!

I'm afraid there's going to be another round of divergence between the TPP and EMIS backends with the dynamic date stuff I'm introducing. But that doesn't directly affect any of the code here.

Issue #345, opened 2020-11-10
inconsistent values for binary variables between dummy and real data


wjchulme:
For example 
``` py
registered = patients.patients.registered_as_of(
        index_date,
        return_expectations={"incidence": 0.5}
    ),
```
will give a column of `"1"`s and `""`s (ie empty) when generating dummy data, but will give a column of `"1"`s and `"0"`s when run on the real data. 

So code that works on dummy data may not work on real data, depending on how these missing values are handled.

PR #384, opened 2020-12-17
fix: Stop pandas converting empty string to None

### LINKED ISSUE #379: Measures drop blanks
When calculating Measures all the people with empty/unknown values for any category/denominator are dropped rather than grouped together into an "unknown" category (e.g. missing ethnicity). While sometimes it is useful to drop these (e.g. if v low numbers) in general we should keep these by default to make sure missing data is as expected. 

Illustrative example from dummy data:
- input.csv:
  image

- same data after grouping to measures (no "blank" category):
  image

evansd:
Closes #379 and depends on #395

### DIFF
--- a/cohortextractor/cohortextractor.py
+++ b/cohortextractor/cohortextractor.py
@@ -321,7 +321,9 @@ def _load_csv_for_measures(file, measures):
     dtype = {col: "category" for col in group_by_columns}
     for col in numeric_columns:
         dtype[col] = "float64"
-    df = pandas.read_csv(file, dtype=dtype, usecols=list(dtype.keys()))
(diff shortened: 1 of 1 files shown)


inglesp (approved):

evansd:
Oh FFS. The test is failing because of issue #345: the dummy data has null values in it where the real data is designed never to output these.

The offending code is this section here:
https://github.com/opensafely/cohort-extractor/blob/ff7667de7ad94c847f851520eb9a28f76f32f20d/cohortextractor/expectation_generators.py#L139-L144

Instead of setting everything to `None` it needs to set the correct "empty" value for each type.

I don't have time to do this now.

PR #385, opened 2020-12-17
Support dynamic date references

### LINKED ISSUE #347: Variable "as of" dates in study definitions
>An additional issue that may need resolving is having more flexibility for "as of" dates. Currently we can ask for example "what's the patient's care home status as of <index_date>", where index date is a scalar fixed across all patients. But we can't ask "what's the patient's care home status as of <hosp_admission_date>", i.e., for vectorised dates.
>
>You'd want this for both one-row-per-patient datasets and one-row-per-event datasets. It would probably need a table of "state transition" dates (care home status, dialysis status, employment status, whatever) to implement
>
>_Originally posted by @wjchulme in https://github.com/opensafely/cohort-extractor/issues/136#issuecomment-726039157_

I think this is distinct and important enough to have its own issue. It's also somewhat urgent, as the lack of this is meaning that some current studies are having to use an awkward and imperfect workaround of defining a variable many times within monthly windows, then piecing those back together subsequently to get a more complete picture of when a thing happened. For example:
 - @StatsFizz's smoking status and bmi in [Risk Prediction](https://github.com/opensafely/risk-prediction-research/blob/db8be510860854c6029cbb472abac4ddba919f8e/analysis/study_definition.py#L243)
 - @angelwong121's drug exposures in [anticoagulants](https://github.com/opensafely/anticoagulants-research/blob/13dcfa0c8ed76eb200dce7cd30d75e26471ac7a3/analysis/common_variables.py#L135)
 - My and @johntaz's outcomes in the thrombosis after COVID ([here](https://github.com/opensafely/post-covid-thrombosis-research) but not yet coded), where we need to determine whether an outcome occurred both before and after the patient index date, but that index date is not fixed (it's the date when a patient was discharged from hospital).
 - Any of the other near future outcomes after COVID study.
 - Any vaccine follow-up studies (which we might have to implement quickly when asked by external people)

While this could be determined using a time varying separate dataset for these variables as described in #136, it's then extracting more data than needed, and is more awkward to code downstream then just being able to do e.g. `on_or_after=list_of_individual_patient_index_dates`. 

cc @DarthCTR and @sebbacon as you also commented on #136.

evansd:
Aka "dynamic index dates", aka "vectorised dates"

This allows you to use the name of another column as a date reference, anywhere where you would use a date e.g
```py
study = StudyDefinition(
    population=patients.all(),
    earliest_frobble=patients.with_these_clinical_events(
        codelist(["frobble"], system="ctv3"),
        returning="date",
        date_format="YYYY-MM-DD",
    ),
    latest_wibble_in_month_after_frobble=patients.with_these_clinical_events(
        codelist(["wibble"], system="ctv3"),
        returning="date",
        date_format="YYYY-MM-DD",
        between=["earliest_frobble", "last_day_of_month(earliest_frobble) + 1 month",
    ),
)
```

**Note**: this does not work in `expectations` definitions. That will need to be updated separately.

Closes #347

----

To properly test this we'd need to go over every single test case in the suite and construct a new test which uses dynamic index dates and `ELITERALLYCANNOTFACEIT` right now. I'm satisfied via the current test suite that I haven't broken existing functionality. And that the dynamic dates work with the clinical events table, which is the main thing people want to use it with. The kind of error I'm expecting is blowing up with an "ambiguous column ref" error when you try to use a dynamic date due to the extra join tables. And I can fix these as they arise if necessary.

Longer term, Plans Are Afoot to refactor this all into oblivion in any case.


tests/test_tpp_backend.py
```diff
@@ -2851,3 +2851,63 @@ def test_extracting_at_different_index_dates():
...
+            find_last_match_in_period=True,
+            on_or_before="last_day_of_month(earliest_bar) + 1 month",
+        ),
+    )
+    assert_results(
+        study.to_dicts(),
+        earliest_bar=["2020-03-01"],
+        latest_foo_before_bar=["2020-02-01"],
+        latest_foo_in_month_after_bar=["2020-04-20"],
+    )
```
inglesp, line 2913: It would be good to have a test that demonstrates the behaviour when a date expression is invalid

inglesp (approved):
This is a chunky piece of work, but it looks well-considered and each commit was easy to understand.  I don't have very much to say, except that I think this probably the last feature we can add without a serious rethink about refactoring -- especially since this is TPP only.

PR #387, opened 2020-12-18
Pull image before building in GH Action

inglesp:
This should speed up builds, since it can skip building earlier layers
that haven't changed

### DIFF
--- a/.github/workflows/build_and_publish.yaml
+++ b/.github/workflows/build_and_publish.yaml
@@ -72,6 +72,8 @@ jobs:
     - name: Generate correct value for VERSION file
       run: |
         echo ${{ needs.tag-new-version.outputs.tag }} > cohortextractor/VERSION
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_publish.yaml
```diff
@@ -72,6 +72,8 @@ jobs:
     - name: Generate correct value for VERSION file
       run: |
         echo ${{ needs.tag-new-version.outputs.tag }} > cohortextractor/VERSION
+    - name: Pull image
+      run: docker pull docker-proxy.opensafely.org/opensafely/cohortextractor
```
bloodearnest, line 76: This should be ghcr.io/ rather than the proxy, as its running it gh. We should only need to use the proxy to pull stuff onto the server.

bloodearnest (approved):

PR #388, opened 2020-12-22
Add noddy REPL for Presto

inglesp:


cohortextractor/presto_utils.py
```diff
@@ -1,8 +1,10 @@
+import readline  # noqa
```
evansd, line 1: I guess something magical happens when you import this? A comment would be helpful

evansd (approved):

cohortextractor/presto_utils.py
inglesp: Yes -- it adds readline behaviour to `input()`.

PR #389, opened 2020-12-22
fix: Allow digits in column names

inglesp:
This also removes `-` as an allowed character, which we should never have supported.


evansd (approved):
Thanks!

PR #390, opened 2020-12-31
feat: Support connecting to Presto with a PKCS#12 certificate

inglesp:
Docs at https://github.com/ebmdatalab/datalab-team-manual/pull/9.

PR #391, opened 2021-01-04
fix: Allow dynamic dates to work with dummy data

evansd:
Actually supporting these correctly during dummy data generation is a
much bigger chunk of work, but for now we just want to ignore them so it
doesn't blow up.

### DIFF
--- a/cohortextractor/study_definition.py
+++ b/cohortextractor/study_definition.py
@@ -1,6 +1,7 @@
 import collections
 import copy
 import os
+import re
 
 import pandas as pd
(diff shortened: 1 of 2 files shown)


evansd:
@inglesp Failures seem to be the Presto issue. Is there a particular PR I should rebase against to fix this?

inglesp (approved):

inglesp:
> Failures seem to be the Presto issue. Is there a particular PR I should rebase against to fix this?

You could rebase against #390.

PR #392, opened 2021-01-04
Verify server certs

inglesp:


cohortextractor/presto_utils.py
```diff
@@ -17,13 +18,22 @@ def presto_connection_from_url(url):
...
+
     return ConnectionProxy(conn)
 
 
 def adapt_connection(conn, conn_params):
     """Adapt connection to use passphrase-protected PKCS#12 certificate.
 
-    See instructions in opensafely/accessing-emis-data/ for getting a certificate.
+    For instructions for getting a certificate, see
+    https://ebmdatalab.github.io/datalab-team-manual/opensafely/accessing-emis-data/
```
ghickman, line 36: 💖

certs/providerplus.emishealthinsights.co.uk/get_certs.sh
```diff
@@ -0,0 +1,3 @@
+openssl s_client -showcerts -connect providerplus.emishealthinsights.co.uk:443 2>/dev/null </dev/null | \
```
ghickman, line 1: Could you stick a shebang at the top of this file?  IIRC it'll run in whatever shell you're calling it from otherwise.

We probably want `set -u` (fail a pipeline if any command fails) too.

cohortextractor/presto_utils.py
bloodearnest: Why this change, OOI? I think we want OS related docs in the OS org, private if needed?

bloodearnest: Oh, is this just an explicit url?

bloodearnest (approved):

cohortextractor/presto_utils.py
inglesp: Just an explicit URL.

PR #395, opened 2021-01-04
fix: Generate correct "empty" values in dummy data

### LINKED ISSUE #345: inconsistent values for binary variables between dummy and real data

evansd:
Closes #345

### DIFF
--- a/cohortextractor/expectation_generators.py
+++ b/cohortextractor/expectation_generators.py
@@ -105,6 +105,8 @@ def generate(population, **kwargs):
         category_labels = dict(zip(category_ids, ratios.keys()))
         df = df.replace({"category": category_labels})
         df["category"] = df["category"].astype("category")
+        # Add empty string as a category to support missing values
+        df["category"].cat.add_categories(new_categories=[""], inplace=True)
 
     int_ = kwargs.pop("int", None)
     if int_:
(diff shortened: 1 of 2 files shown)


cohortextractor/expectation_generators.py
```diff
@@ -138,10 +140,21 @@ def generate(population, **kwargs):
...
         df = df.drop("date", axis=1)
     return df
+
+
+def set_empty_values(df, row_selection):
+    empty_values = {
+        "bool": False,
+        "int": 0,
+        "float": 0.0,
+        "date": None,
```
sebbacon, line 157: Is this correct? This suggests date should be empty string:

https://github.com/opensafely/cohort-extractor/blob/c4d1fc741d59e5ae7e2d585e6db880608e01a44c/cohortextractor/emis_backend.py#L173-L185

Or is "default" different from "empty"?

evansd: Ah yes! This is intentional and I meant to add a comment here but forgot. I'll update.

sebbacon (approved):

PR #409, opened 2021-01-07
Retry failed queries

inglesp:
This is all fairly speculative, but might help us characterise the problems

### DIFF
--- a/cohortextractor/presto_utils.py
+++ b/cohortextractor/presto_utils.py
@@ -13,6 +13,9 @@ import urllib3
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
 
+LOG_SQL = bool(os.getenv("LOG_SQL"))
+
+
 def presto_connection_from_url(url):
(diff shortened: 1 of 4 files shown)


cohortextractor/presto_utils.py
```diff
@@ -6,13 +6,21 @@
...
 from requests_pkcs12 import Pkcs12Adapter
+from retry import retry
 from tabulate import tabulate
 
 # TODO remove this when certificate verification reinstated
 import urllib3
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
 
+LOG_SQL = bool(os.getenv("LOG_SQL"))
```
sebbacon, line 17: Should prob either call this `LOG_PRESTO_SQL` or also add it to the TPP backend?

inglesp: This is all speculative, we can do it right later.

sebbacon (approved):

PR #410, opened 2021-01-08
Add linting etc

inglesp:


evansd:
We have configured, but not propitiated, Pendantic Pull Request Check

cohortextractor/emis_backend.py
```diff
@@ -541,7 +543,10 @@ def patients_with_these_medications(self, **kwargs):
             # Remove unhandled arguments and check they are unused
             assert not kwargs.pop("episode_defined_as", None)
             return self._patients_with_events(
-                MEDICATION_TABLE, "", "snomed_concept_id", **kwargs,
```
ghickman, line 544: If you're keen to keep this line (and other similar ones) as close to the original as before then you could remove the final comma instead.  A recent black release changed trailing commas to force one item per line.

inglesp: Thanks.  I'm not really fussed tbh.

.github/workflows/test_runner.yaml
```diff
@@ -1,40 +1,82 @@
...
+        with:
+          python-version: "3.7"
+      - uses: actions/cache@v2
+        with:
+          path: ${{ github.workspace }}/venv
+          key: venv-${{ github.ref }}-${{ hashFiles('requirements.txt') }}
+      - name: Run all pre-commit checks
+        run: |
+          source ${{ github.workspace }}/venv/bin/activate
+          pre-commit run --all-files
```
ghickman, line 47: This is pretty neat, but the downside is local tooling can now differ from what's run in CI because dependabot can bump the requirements in requirements.txt but not .pre-commit-config.yaml.  I'm not sure that will actually cause a problem yet, just flagging it!

ghickman (changes requested):
Looking good!

One nit: you'll need to exclude common directories from black like you've done with flake8.

.github/workflows/test_runner.yaml
inglesp: Yeah -- I began to look into getting pre-commit to use the cached virtualenv, but ran out of steam quite quickly!

I think this is a strict improvement on what we have now, but it's something to be aware of.

ghickman: > strict improvement on what we have now

Big +1

I've tried the cached virtualenv thing in the past and also gave up 😬

I'm curious to see how this flow works out 🙂

inglesp:
> One nit: you'll need to exclude common directories from black like you've done with flake8.

Looks like black respects .gitignore, so I don't think this will be necessary.

ghickman:
> black respects .gitignore

Oh perfect, ignore me then 🙂

ghickman (approved):

PR #415, opened 2021-01-11
fix: Support using dynamic dates with SGSS queries

evansd:


inglesp (approved):
Greenspun's tenth rule, but for ORMs.

PR #417, opened 2021-01-11
Add logging

inglesp:
This PR:

* allows us to log SQL sent to the presto backend (by setting `LOG_SQL`)
* allows us to get debug logs from underlying libraries (by setting `LOG_LEVEL`)
* changes from `print()` to `logger.info()` for presenting information to users, which means that the output is now slightly noisier -- we may want to revisit this

### DIFF
--- a/cohortextractor/cohortextractor.py
+++ b/cohortextractor/cohortextractor.py
@@ -21,6 +21,7 @@ import numpy as np
 import pandas
 import requests
 import seaborn as sns
+import structlog
 import yaml
 from matplotlib import pyplot as plt
 from pandas.api.types import (
@@ -33,6 +34,8 @@ from prettytable import PrettyTable
 
 import cohortextractor
 from cohortextractor.localrun import localrun
+
+logger = structlog.get_logger()
 
 notebook_tag = "opencorona-research"
 target_dir = "/home/app/notebook"
@@ -139,7 +142,6 @@ def generate_cohort(
                 study_definitions = [(study_name, suffix)]
                 break
     for study_name, suffix in study_definitions:
-        print(f"Generating cohort for {study_name}...")
         _generate_cohort(
             output_dir,
             study_name,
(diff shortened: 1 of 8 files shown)


inglesp:
Thanks to @ghickman for the earlier logging seminar.

cohortextractor/cohortextractor.py
```diff
@@ -158,7 +160,17 @@ def _generate_cohort(
     index_date_range=None,
     skip_existing=False,
 ):
-    print("Running. Please wait...")
+    logger.info(
+        "Generating cohort for {study_name} in {output_dir}",
```
ghickman, line 164: Missing `f` string prefix here?

Side note: stdlib logging is [optimised](https://docs.python.org/3/howto/logging.html#optimization) when using `%` style formats.  I suspect we'll never see the benefit of it here but thought you could use just a smidge more logging info for the day!

evansd (approved):
This looks good (though even with structlog, Python logging still manages to make me feel sad)

cohortextractor/cohortextractor.py
inglesp: Well spotted!

bloodearnest: Using f strings is fine here IMO - it's clear, consistent, and negligible performance overhead is not a problem in this scenario.

FWIW, it's not %-style per se (you can use {} style if you've set a logging.Formatter with style='{') - it's the delayed string interpolation which is the optimisation.

e.g. If you do:

    logger.info('%s', 'msg')
    # or with style='{'
    logger.info('{}', 'msg')

Then the string will only be interpolated iff the log message is actually emitted by a handler. i.e. if it's a DEBUG level log, and you're only logging INFO, the interpolation will be skipped. Can be useful if you have a lot of DEBUG level logs with logs of embedded string interpolation - avoid the costs if not turned on.

PR #425, opened 2021-01-15
feat: Add minimal support for the High Cost Drugs dataset

evansd:


tests/test_tpp_backend.py
```diff
@@ -2983,3 +2985,51 @@ def test_dynamic_index_dates_with_invalid_expression():
...
+            find_last_match_in_period=True,
+            returning="date",
+            date_format="YYYY-MM",
+        ),
+        latest_bar_date=patients.with_high_cost_drugs(
+            drug_name_matches="bar",
+            find_last_match_in_period=True,
+            returning="date",
+            date_format="YYYY-MM",
+        ),
```
inglesp, line 3022: This is fine but since there is only one `bar` event, the test will pass if you change `find_last_match_in_period` to `find_first_match_in_period`.

inglesp (approved):
Laughing like a drain

```
DATEFROMPARTS(
  (CAST(FinancialYear AS INT) / 100) + IIF(CAST(FinancialMonth AS INT) > 9, 1, 0),
  ((CAST(FinancialMonth AS INT) + 2) % 12) + 1,
  1
) AS date
```

tests/test_tpp_backend.py
evansd: Good point. I've made the tests slightly more extensive.

PR #429, opened 2021-01-18
Update Dockerfile and switch to Python 3.8

### LINKED ISSUE #419: We should use the same base docker image as other OpenSAFELY projects
And we should apt install Python rather than compiling it with `pyenv` as we do here:
https://github.com/opensafely/job-runner/blob/master/Dockerfile

evansd:
This updates to Dockerfile to use the [base-docker][1] image, use pre-compiled Python binaries from apt rather than compiling using pyenv, and switches to Python 3.8.

Closes #419

[1]: https://github.com/opensafely-core/base-docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,64 +1,39 @@
-FROM ubuntu:bionic
+FROM ghcr.io/opensafely-core/base-docker
 
-# PYTHON_VERSION can be changed, by passing `--build-arg
-# PYTHON_VERSION=<new version>` during docker build
-ARG PYTHON_VERSION=3.7.8
-ENV PYTHON_VERSION=${PYTHON_VERSION}
+RUN \
+  apt-get update --fix-missing && \
+  apt-get install -y python3.8 python3.8-dev python3-pip git curl unixodbc-dev && \
+  update-alternatives --install /usr/bin/python python /usr/bin/python3.8 1
 
-ENV DEBIAN_FRONTEND noninteractive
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,64 +1,39 @@
...
+WORKDIR /app
+
+# We run `cohortextractor --help` at the end to force dependencies to import
+# because the first time we import matplotlib we get a "generated new
+# fontManager" message and we want to trigger that now rather than every time
+# we run the docker image
+RUN \
+  python -m pip install --upgrade 'pip>=21,<22' && \
+  python -m pip install --requirement requirements.txt && \
+  cohortextractor --help
```
inglesp, line 36: Oh computer

requirements.txt
```diff
@@ -52,13 +52,6 @@ identify==1.5.11
     # via pre-commit
 idna==2.9
     # via requests
-importlib-metadata==3.3.0
-    # via
-    #   flake8
-    #   pluggy
-    #   pre-commit
-    #   pytest
-    #   virtualenv
```
inglesp, line 61: Do these changes matter?

inglesp (approved):
Nice, lots of help for future-us.

requirements.txt
evansd: No, `importlib-metadata` is a backport of 3.8 stdlib functions which was needed on 3.7 but isn't on 3.8. (I didn't intentionally remove it, this was just pip-compile doing its thing.)

bloodearnest (approved):

PR #439, opened 2021-01-28
feat: Support querying clinical events with SNOMED codelists

inglesp:


cohortextractor/tpp_backend.py
```diff
@@ -1047,10 +1059,10 @@ def _number_of_episodes_by_clinical_event(
             codelist, case_sensitive=True
         )
         date_condition, date_joins = self.get_date_condition(
-            "CodedEvent", "ConsultationDate", between
+            from_table, "ConsultationDate", between
         )
         ignored_day_condition, extra_queries = self._these_codes_occur_on_same_day(
-            "CodedEvent", ignore_days_where_these_codes_occur, between
+            from_table, ignore_days_where_these_codes_occur, between
         )
```
evansd, line 1066: Github won't let me comment on it directly because it's not part of the diff, but the `_these_codes_occur_on_same_day()` function has references that need fixing too.

evansd (changes requested):
This is great. Not as many changes as I'd feared. There is one function that still needs amending though.

evansd (approved):

PR #442, opened 2021-01-29
fix: Remove push to old Docker registry

evansd:
For now, we need to manually pull newly built images from `opensafely-core` and push them to `opensafely` so they can be used with the job-runner:
```
docker pull ghcr.io/opensafely-core/cohortextractor:latest
docker tag ghcr.io/opensafely-core/cohortextractor:latest ghcr.io/opensafely/cohortextractor:latest
docker push ghcr.io/opensafely/cohortextractor:latest
```


inglesp (approved):
Thanks for fixing!

PR #444, opened 2021-01-29
fix: Support returning SNOMED code from with_these_events query

inglesp:


evansd (approved):
Ahhh! Yes, I see the issue

PR #449, opened 2021-02-01
Support vaccine studies in EMIS

inglesp:


cohortextractor/emis_backend.py
```diff
@@ -1165,6 +1165,10 @@ def quote(value):
         return str(value)
 
     value = str(value)
+
+    if re.match(r"^\w+ \+ interval '\d+' (day|month|year)$", value):
+        return value
+
```
evansd, line 1173: I don't know enough of the context here to comment really. I'm a bit surprised this all works but I'll trust you that it does.

Do you need to support negative intervals?

Is the idea that this is temporary and we'll move to the existing `<date> + 10 days` syntax at some point?

evansd (approved):
Happy to approve this but I don't fully understand what the date parsing changes are doing

cohortextractor/emis_backend.py
inglesp: Was sure I'd written an explanation / note of apology somewhere, sorry.

Yes, the idea is this is temporary.  Supporting `<date> + 10 days` will presumably come for free (with the code to actually generate the SQL as a rounding error) once we've done the big refactor.

evansd: :+1: Makes sense

PR #453, opened 2021-02-05
fix: Escape characters rather than using a safelist

evansd:
This is prompted by the need to query the High Cost Drugs data in which
drug names are (apparently) entered as free text. This leads to names
such as:

    infliximab-remicade!100mg!infusion
    infliximab 'remicade' (10ml vi
    infliximab 'remicade' (10ml vial)
    ixekizumabß

Given the presence of single quotes in these names the only way we can
use them in queries is to escape them properly.

We also take this opportunity to correctly escape wildcard characters in
strings which are going to be used in LIKE queries.

### DIFF
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -17,11 +17,6 @@ from .mssql_utils import (
 
 logger = structlog.get_logger()
 
-# Characters that are safe to interpolate into SQL (see
-# `placeholders_and_params` below)
-safe_punctation = r" _.-+/()"
-SAFE_CHARS_RE = re.compile(f"^[a-zA-Z0-9{re.escape(safe_punctation)}]+$")
-
 
 class TPPBackend:
     _db_connection = None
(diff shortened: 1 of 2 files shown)


cohortextractor/tpp_backend.py
```diff
@@ -1992,24 +1987,32 @@ def __str__(self):
...
-        patterns.append(f"'{prefix}{quoted_code[1:-1]}{suffix}'")
+    for code in codelist_to_list(codelist):
+        escaped_code = escape_like_query_fragment(code)
+        patterns.append(quote(f"{prefix}{escaped_code}{suffix}"))
     return patterns
 
 
+# Special characters from:
+# https://docs.microsoft.com/en-us/sql/t-sql/language-elements/like-transact-sql?view=sql-server-ver15
+LIKE_ESCAPE_TABLE = str.maketrans({char: f"\\{char}" for char in "%_[]^"})
```
inglesp, line 2007: A rare sighting of `str.maketrans` in the wild!

inglesp (approved):
Looks good.  It might be nice to have tests that demonstrate that the escaping actually works by hitting the database, but only if easy.

cohortextractor/tpp_backend.py
evansd: I was halfway through writing a hideous `re.sub` expression with three nested layers of backslash escaping when I thought to myself "there must be a better way". And there was!

evansd:
> It might be nice to have tests that demonstrate that the escaping actually works by hitting the database

Good suggestion, have done so.

inglesp:
> Good suggestion, have done so.

Nice tests.

PR #461, opened 2021-02-11
feat: Add numeric_value_in_range

### LINKED ISSUE #458: Support ignoring null numeric_values
To implement the PRIMIS vaccine spec I need to find the date of the latest observation with a non-null value.

I propose updating the API to allow:

```
patients.with_these_clinical_events(
    codelist,
    returning="last_date_in_period",
    ignore_null_values=True,
    ...
)
```

`ignore_null_values` would be valid for any value of `returning`, and would add `AND NumericValue IS NOT NULL` to the relevant SQL in `_patients_with_events`.

inglesp:
Fixes #458

I'm having second thoughts about whether this approach is better than adding `ignore_null_values`, with a note in the docs that NULLs are converted to zero in the TPP backend.  Thoughts?

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -513,6 +513,7 @@ def with_these_clinical_events(
     returning="binary_flag",
     include_date_of_match=False,
     date_format=None,
+    numeric_value_in_range=(None, None),
     # Special (and probably temporary) arguments to support queries we need
     # to do right now. This API will need to be thought through properly at
     # some stage.
@@ -565,6 +566,9 @@ def with_these_clinical_events(
             ignored. if a events is found on this day, the date is not matched even it matches a
             code in the main `codelist`
         episode_defined_as: a string expression indicating how an episode should be defined
+        numeric_value_in_range: a pair of numbers `(lower, upper)`.  Only events whose
+            numeric value is between `lower` and `upper` are matched.  Either `lower` or
+            `upper` may be `None`.
         return_binary_flag: a boolean indicating if the number of matches in a period should be
             returned (deprecated: use `date_format` instead),
         return_number_of_matches_in_period: a boolean indicating if the number of matches in a period should be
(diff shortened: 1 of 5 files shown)


cohortextractor/patients.py
```diff
@@ -565,6 +566,9 @@ def with_these_clinical_events(
             ignored. if a events is found on this day, the date is not matched even it matches a
             code in the main `codelist`
         episode_defined_as: a string expression indicating how an episode should be defined
+        numeric_value_in_range: a pair of numbers `(lower, upper)`.  Only events whose
+            numeric value is between `lower` and `upper` are matched.  Either `lower` or
+            `upper` may be `None`.
```
ghickman, line 571: Should we include whether they're inclusive/exclusive explicity here?  It feels like something would be useful to call out?

evansd:
I don't see anything wrong with this PR, but I do share your feeling that maybe the simpler, less powerful thing is actually better here. I wonder if something like `ignore_missing_values` or `ignore_empty_values` might better convey the uncertainty whereas "null" communicates something very specific which isn't quite true.

I'm just going to double check my understanding of this against the TPP database; but even if it turns out that they do now support NULLs I still think we should probably be ignoring zero values as well (assuming they don't make sense in the clinical context) because we don't know how this data was recorded historically.

evansd:
OK, have confirmed that `NumericValue` is still non-nullable in TPP.

inglesp:
OK thanks.  Tet's do `ignore_missing_values`, with a note to say that missing values are recorded as zero in TPP.

evansd:
Do you think it's worth ignoring zeros in EMIS as well?

inglesp:
> Do you think it's worth ignoring zeros in EMIS as well?

Zero can be a sensible value -- eg of alcohol intake.

I suppose it's probably better to have consistency across backends, and then let study authors work out what how to handle zeros if they need to.

sebbacon:
I think the latest proposal is the most expressive, and also addresses the user need most directly.  Values between X and Y are arguably better suited in analytic scripts.

It makes me think we probably want a warnings framework to hint to users at run time of oddities / inconsistencies like this

inglesp:
Closing in favour of #469.

Issue #436, opened 2021-01-22
Add SGSS SGTF/channels data


HelenCEBM:
When available, will be needed ASAP for a range of studies.
cc @rozeggo

rozeggo:
Paired with #437 for updating.

rozeggo:
From Chris_TPP:
New variant SGSS is in OpenSAFELY . @Jonny has just built it. One more column called SGTF, just on the positive tests table. Not on the all tests table. Column is varchar for safety but has values blank, 0, 1, and 9.

PR #463, opened 2021-02-12
Support new SGSS tables and SGTF field

### LINKED ISSUE #437: Add newer SGSS data feeds to study definition
Currently we use the first-test-result-per-patient data but we should add the option to use the newer data which we think has all test results per patient. Important for ongoing work.

evansd:
The updated docstring in `patients.py` contains most of the detail here.

It's now possible to access the "AllTests" data (as opposed to the "Earliest Specimen Date" data) by passing the flag:
`restrict_to_earliest_specimen_date = False` e.g

```py
last_positive_test_date=patients.with_test_result_in_sgss(
    pathogen="SARS-CoV-2",
    test_result="positive",
    restrict_to_earliest_specimen_date=False,
    returning="date",
    date_format="YYYY-MM-DD",
    return_expectations={
        "date": {"earliest" : "2020-02-01"},
        "rate" : "exponential_increase"
    },
)
```

S-gene target failure results can be obtained using `returning = "s_gene_target_failure"` e.g.
```py
sgtf_result=patients.with_test_result_in_sgss(
    pathogen="SARS-CoV-2",
    test_result="positive",
    restrict_to_earliest_specimen_date=True,
    returning="s_gene_target_failure",
    return_expectations={
        "rate": "universal",
        "category": {
            "ratios": {
                "": 0.7,
                "0": 0.1,
                "1": 0.1,
                "9": 0.1,
            },
        },
    },

)
```

Note it's **not** possible to combine both the "AllTests" and "SGTF" options because the underlying data doesn't provide this. The cohortextractor will shout loudly at you if you try.

Closes #437 and #436

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -1336,25 +1336,30 @@ def with_test_result_in_sgss(
     # Matching rule
     find_first_match_in_period=None,
     find_last_match_in_period=None,
+    restrict_to_earliest_specimen_date=True,
     # Set return type
     returning="binary_flag",
     date_format=None,
     return_expectations=None,
 ):
     """
-    Finds lab test results recorded in SGSS (Second Generation Surveillance
-    System).
+    Finds COVID lab test results recorded in SGSS (Second Generation
+    Surveillance System).
 
-    Please note for the dates this is used in the database as the date the specimen was taken, rather than the
-    date of the lab result.
+    Please note that all dates used here are "specimen dates" (i.e. the date
+    the specimen was taken), rather than the date the lab result was obtained.
 
-    There's an important caveat here: where a patient has multiple positive
-    tests, SGSS groups these into "episodes" (referred to as
-    "Organism-Patient-Illness-Episodes"). Each pathogen has a maximum episode
-    duration (usually 2 weeks) and unless positive tests are separated by
-    longer than this period they are assumed to be the same episode of illness.
-    The specimen date recorded is the *earliest* positive specimen within the
-    episode.
+    It's important to note that data is supplied in two separate datasets: an
+    "Earliest Specimen" dataset and an "All Tests" dataset.
+
+    #  Earliest Specimen Dataset
+
+    Where a patient has multiple positive tests, SGSS groups these into
+    "episodes" (referred to as "Organism-Patient-Illness-Episodes"). Each
+    pathogen has a maximum episode duration (usually 2 weeks) and unless
+    positive tests are separated by longer than this period they are assumed to
+    be the same episode of illness.  The specimen date recorded is the
+    *earliest* positive specimen within the episode.
 
     For SARS-CoV-2 the episode length has been set to infinity, meaning that
     once a patient has tested positive every positive test will be part of the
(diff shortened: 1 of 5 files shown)


cohortextractor/patients.py
```diff
@@ -1378,39 +1382,91 @@ def with_test_result_in_sgss(
...
+      0: S gene detected
+      Detectable S gene (CH3>0)
+      Detectable y ORF1ab CT value (CH1) <=30 and >0
+      Detectable N gene CT value (CH2) <=30 and >0
+
+      9: Cannot be classified
+
+      Null are where the target is not S Gene. I think LFTs are currently
+      also coming across as 9 so will need to review those to null as well as
+      clearly this is a PCR only variable.
```
sebbacon, line 1418: Does this snippet of info need actioning? Either way I'm not sure it'll be useful to end-users in our documentation as it is?

cohortextractor/patients.py
```diff
@@ -1378,39 +1382,91 @@ def with_test_result_in_sgss(
...
+
+    For more detail on SGSS in general see:
+    https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/739854/PHE_Laboratory_Reporting_Guidelines.pdf
```
sebbacon, line 1421: As this will be pulled into docs, should this be a markdown link?

cohortextractor/patients.py
```diff
@@ -1336,25 +1336,30 @@ def with_test_result_in_sgss(
...
 
-    There's an important caveat here: where a patient has multiple positive
-    tests, SGSS groups these into "episodes" (referred to as
-    "Organism-Patient-Illness-Episodes"). Each pathogen has a maximum episode
-    duration (usually 2 weeks) and unless positive tests are separated by
-    longer than this period they are assumed to be the same episode of illness.
-    The specimen date recorded is the *earliest* positive specimen within the
-    episode.
+    It's important to note that data is supplied in two separate datasets: an
+    "Earliest Specimen" dataset and an "All Tests" dataset.
```
sebbacon, line 1353: I think the exposition of why we're even explaining this could be improved here. It's awkward to explain, but how about something like:

> It's important to note that due to limitations of the underlying raw data, you will not be able to return both SGTF results (`returning='s_gene_target_failure'`) and query all positive test results for an individual (`restrict_to_earliest_specimen_date=False`) in a single call. The following explains why.

cohortextractor/patients.py
```diff
@@ -1378,39 +1382,91 @@ def with_test_result_in_sgss(
...
+        between: two dates of interest as a list with each date as a string
+            with the format `YYYY-MM-DD`.  Filters results to between the two
+            dates provided. The default value is `None`.
+        find_first_match_in_period: a boolean that indicates if the data
+            returned is first event if there are multiple matches within the
+            time period
+        find_last_match_in_period: a boolean that indicates if the data
+            returned is last event if there are multiple matches within the
+            time period
+        restrict_to_earliest_specimen_date: a boolean indicating whether to use
```
sebbacon, line 1443: I find this a bit awkward, specially when describing it in the "exposition" bit above.

Would `treat_every_specimen_as_new_episode` be any better. Just spitballin

ghickman (approved):
Great docstring 👍

evansd:
@sebbacon Agree with all your comments, but someone other than me can rewrite the docs. I'm extremely fed up with the woeful and utterly undocumented state SGSS is provided to us in.

PR #469, opened 2021-02-16
feat: Add ignore_missing_values option for with_these_clinical_events()

### LINKED ISSUE #458: Support ignoring null numeric_values

inglesp:
Fixes #458.

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -513,6 +513,7 @@ def with_these_clinical_events(
     returning="binary_flag",
     include_date_of_match=False,
     date_format=None,
+    ignore_missing_values=False,
     # Special (and probably temporary) arguments to support queries we need
     # to do right now. This API will need to be thought through properly at
     # some stage.
@@ -565,6 +566,9 @@ def with_these_clinical_events(
             ignored. if a events is found on this day, the date is not matched even it matches a
             code in the main `codelist`
         episode_defined_as: a string expression indicating how an episode should be defined
+        ignore_missing_values: ignore events where the value is missing.  Note that in
+            TPP, a missing value is recorded as a zero, so we do not distinguish between
+            values that are zero and values are missing in either backend
         return_binary_flag: a boolean indicating if the number of matches in a period should be
             returned (deprecated: use `date_format` instead),
         return_number_of_matches_in_period: a boolean indicating if the number of matches in a period should be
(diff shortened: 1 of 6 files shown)


evansd (approved):
Nice

cohortextractor/patients.py
```diff
@@ -565,6 +566,9 @@ def with_these_clinical_events(
             ignored. if a events is found on this day, the date is not matched even it matches a
             code in the main `codelist`
         episode_defined_as: a string expression indicating how an episode should be defined
+        ignore_missing_values: ignore events where the value is missing.  Note that in
+            TPP, a missing value is recorded as a zero, so we do not distinguish between
+            values that are zero and values are missing in either backend
```
sebbacon, line 571: Late to the party on this one (sorry!) but I don't quite understand that last sentence?

inglesp: In TPP, the value field is not nullable, so missing values are recorded as a zero.  In EMIS, the value field is nullable.  Until we can give researchers access to row-level data, we thought it better to give consistency between backends.  (See #461)

sebbacon: I'm sorry, I wasn't clear. What I mean is, I know what the underlying situation is, but the way it's expressed here to a new user probably won't make sense?  I presume you mean something like this?

```
ignore_missing_values: ignore events where the value is missing.  Note that in
            TPP, a missing value is recorded as a zero, so this only has an effect in the EMIS backend
```

inglesp: This isn't correct -- it has an effect in both backends but I can see the confusion.

If `ignore_missing_values` is set then in TPP we ignore records whose value is zero, and in EMIS we ignore records whose value is zero or null.

So how about adding "or zero":

```
 ignore_missing_values: ignore events where the value is missing or zero.  Note that in
            TPP, a missing value is recorded as a zero, so we do not distinguish between
            values that are zero and values are missing in either backend
```

?

sebbacon: I think this is clearer?
```
ignore_missing_values: ignore events where the value is missing or zero.  Note that in
            TPP, a missing value is recorded as a zero, so in TPP this ignores records that are zero and EMIS this ignores records whose value is zero or null.
```

inglesp: That's great, thanks.

PR #473, opened 2021-02-23
Remove .python-version files and its docker volume

ghickman:
Now that we're not using pyenv in the Docker image I think this bit of docker-compose config can go away.

The current way can cause issues with pyenv when you don't have a local .python-version but are still using pyenv (I'm configuring my venv and python version via direnv), once the app container is stood up and `.python-version-docker` is copied in an empty copy appears on my local FS (I think that's because of the other volume mount?) and pyenv gets very confused because I don't have a python 3.8.1 (I'm on 3.8.7).

So I _think_ this is safe to remove because a local `.python-version` should no longer be an issue for local testing?


evansd (approved):
Yes, good catch. I think the semantic check is failing because, annoyingly, you have to have the `fix:` prefix in the commit title not just the PR title. You could probably get away with something like `chore:` here though as we don't actually need to create a new release for this (right?)

PR #478, opened 2021-02-24
Support date arithmetic in EMIS backend

inglesp:
Have fun @evansd!


cohortextractor/emis_backend.py
```diff
@@ -70,7 +71,7 @@ def to_dicts(self):
         output = [dict(zip(keys, map(str, row))) for row in result]
         unique_check = UniqueCheck()
         for item in output:
-            unique_check.add(item["patient_id"])
+            unique_check.add(item["registration_id"])
```
evansd, line 74: There may well be good reasons for doing this, but as an initial reaction the fact that EMIS uses `registration_id` internally feels like a bit of an implementation detail that shouldn't leak out here.

evansd (approved):
I suspect that when Seb said the diff was easy to review he had missed the "Large diffs are hidden by default" section.

Nevertheless, I do think this looks largely correct. The most likely space for bugs (in my experience) is failing to qualify column references in SQL fragments such that they work OK until you add an extra join and then they blow up as ambiguous. I didn't spot any of these issues, but then I missed several when I wrote the TPP version of this myself. They're generally easy enough to fix and result in a loud error rather than silently incorrect results so all-in-all we probably don't need to worry too much about them.

cohortextractor/emis_backend.py
inglesp: Have reverted in 45f442f2aafa23451399b6c6ef46cacfd2ff2885.

inglesp:
@evansd please can you review the latest commits?

cohortextractor/emis_backend.py
```diff
@@ -20,7 +20,26 @@
...
+#     SELECT *
+#     FROM patient_all_orgs_v2
+#     WHERE registration_id NOT IN (
+#         SELECT registration_id
+#         FROM patient_all_orgs_v2
+#         GROUP BY registration_id
+#         HAVING COUNT(*) > 1
+#     )
+# )
+PATIENT_TABLE = "patient_no_duplicates"
```
evansd, line 42: If I understand correctly this is ignoring all registration IDs which occur more than once, rather than choosing one registration arbitrarily. I'm assuming that the number of duplicates is small enough that it doesn't matter that it causes us to lose a few patients, but just wanted to flag that up. (Is it worth recording in the comment?)

cohortextractor/emis_backend.py
```diff
@@ -78,8 +78,9 @@ def to_csv(self, filename):
         with open(filename, "w", newline="") as csvfile:
             writer = csv.writer(csvfile)
             writer.writerow([x[0] for x in result.description])
-            for row in result:
+            for ix, row in enumerate(result):
                 unique_check.add(row[0])
+                row[0] = ix
```
evansd, line 83: I think this is sensible, but I know there have been a couple of occasions where researchers have relied on patient IDs being consistent to join across extracts from multiple study definitions. The sequential ints here mean that if they try to do this the join will "work" in the sense that there'll be lots of matches, but the results will be nonsense. I wonder if generating a large random offset would help avoid this. Or is consistency between multiple extracts of the same study is desirable we could base the offset on a hash of the SQL or something like that.

Probably not worth worrying about in the immediate term, but we might need to revisit this at some point.

evansd (approved):

cohortextractor/emis_backend.py
inglesp: Thanks, have updated.

cohortextractor/emis_backend.py
inglesp: Maybe we could do `row[0] = md5(row[0])` or something.  (@bloodearnest's original idea.)

PR #480, opened 2021-02-25
Add Ethnicity from SUS

### LINKED ISSUE #270: get ethnicity from APCS
As discussed here https://github.com/opensafely/cohort-extractor/issues/186#issuecomment-665717565

I think as a single-purpose function for now, something like 

    patients.admitted_to_hospital_ethnicity(
      find_most_recent_nonmissing_in_period=True, # or maybe this should always happen so not an option
      return_null_if_conflict = True, # if there are at least two non-null values that don't match
      on_or_before=date,  # optional
      on_or_after=date,  # optional
      between=date,  # optional
    )

In the unlikely event that there are conflicting ethnicity values across different admissions for the same patient, then the `find_most_recent_nonmissing_in_period` and `return_null_if_conflict ` options allow us to resolve it predictably. @rohinimathur let me know if you need it to work differently. 

dictionary for ethnicity codes in APCS to follow shortly

ghickman:
This adds `patient_with_ethnicity_from_sus()` to allow:

```python
patients.with_ethnicity_from_sus(
    returning="code",  # group_6 and group_16 also available
)
```

Fixes #270

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -1923,5 +1923,28 @@ def with_high_cost_drugs(
     return "with_high_cost_drugs", locals()
 
 
-def ethnicity_from_sus(*, setting, on_conflict):
-    pass
+def with_ethnicity_from_sus(
+    returning=None,
+    use_most_frequent_code=None,
+    return_expectations=None,
+):
+    """
+    Returns ethnicity data from the SUS Datasets
+
+    Args:
+        returning: a string indicating what type of value whould be returned.
+        Options are:
+            code: don't group ethnicities at all, return the recorded code
+            group_6: group ethnicities into 6 groups
+            group_16: group ethnicities into 16 groups
+        use_most_frequent_code: when multiple codes are present, pick the most
+        frequent one
+        return_expectations: a dictionary describing what dummy data should
+        look like
+
+    Example:
+        ethnicity_by_16_grouping=patients.with_ethnicity_from_sus(
+            returning="group_16"
+        )
+    """
+    return "with_ethnicity_from_sus", locals()
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -1921,3 +1921,7 @@ def with_high_cost_drugs(
     if date_format == "YYYY-MM-DD":
         raise ValueError("Day level data not available for high cost drugs")
     return "with_high_cost_drugs", locals()
+
+
+def ethnicity_from_sus(*, setting, on_conflict):
+    pass
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -2032,6 +2032,129 @@ class TPPBackend:
         GROUP BY t.Patient_ID
         """
 
+    def patients_with_ethnicity_from_sus(
+        self,
+        returning="code",
+        use_most_frequent_code=None,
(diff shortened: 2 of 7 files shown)


ghickman:
Currently the general flow has been to union the three tables together in a subselect, filtering out the known unwanted data, then build a colum we can use outside the CTE (more on this below) to pick one ethnicity code, with optional mapping to our 6 or 16 groupings.

Picking an ethnicity code per patient is currently done by picking the latest record for a given patient.  I looked at using the suggested `PERCENTILE_*` functions but we can't use those due to the database version.  I'm expecting to revist this after getting the other bits working.

cohortextractor/patients.py
```diff
@@ -1923,5 +1923,19 @@ def with_high_cost_drugs(
     return "with_high_cost_drugs", locals()
 
 
-def ethnicity_from_sus(*, setting, on_conflict):
-    pass
+def with_ethnicity_from_sus(returning=None):
+    """
+    Returns ethnicity data from the SUS Datasets
+
+    Args:
```
evansd, line 1930: I think this needs an extra keyword arg to describe the matching logic. Something like:
```
  use_most_recent_code=None
```
This wouldn't _do_ anything other than we'd have an `assert use_most_recent_code` somewhere to make sure that it was being passed as an argument. This means that's it's hopefully obvious to anyone reading the study def what the logic is. And it future-proofs us for when we add `use_most_frequent_code`.

cohortextractor/tpp_backend.py
```diff
@@ -2032,6 +2032,122 @@ def patients_with_high_cost_drugs(
         GROUP BY t.Patient_ID
         """
 
+    def patients_with_ethnicity_from_sus(self, returning="code"):
+        returning_options = ["code", "group_6", "group_16"]
+        if returning is not None and returning not in returning_options:
```
evansd, line 2037: How come `returning` is allowed to be `None` here?

evansd (changes requested):
This is great. Anyone would think you'd been writing cohortextractor functions all your life! Sorry it's all so painful.

The only thing we have to change is accepting (but not doing anything with) the `return_expectations` argument in the `patients.py` function. I've made a couple of smaller suggestions as well.

Other than those I think this is ready to go.

cohortextractor/tpp_backend.py
ghickman: Good question!  I _think_ I was planning to return either of our two groups alongside the raw code initially but then discussions made me realise we should always let the StudyDefinition author pick, I'll remove the None check as I think authors should always pick what they want… and that's how it's been coded.

ghickman: Fixed and rebased.

cohortextractor/patients.py
ghickman: Nice.  I think you got the names the wrong way round?  But if anything that just shows why it's good 😁

Added, documented, and rebased in.

ghickman:
@evansd – thank you!

I've addressed all your points, could you give them another check for me?

cohortextractor/patients.py
evansd: Ah, yes! Exactly!

evansd (approved):

PR #481, opened 2021-02-25
fix: Allow use of default_expectations for non-date fields

inglesp:
As far as I can tell, there's no reason not to use the `default_expectations` for non-date fields.


sebbacon (approved):

evansd (approved):
As best as I can tell, this is a sensible change

bloodearnest (approved):

PR #482, opened 2021-02-26
fix: add support for plain text cert/key in env var

bloodearnest:
Env vars are PRESTO_TLS_KEY and PRESTO_TLS_CERT

Encrypted pkcs files still supported for now, although we probably want
to drop it at some point

Tried to write a test for the pkcs stuff, but OpenSSL (which
requests_pkcs uses) didn't like the test cert I generated with
cryptography for some reason. Given we'll probably drop it, I figured it
wasn't worth it.

### DIFF
--- a/cohortextractor/presto_utils.py
+++ b/cohortextractor/presto_utils.py
@@ -1,5 +1,6 @@
 import os
 import readline  # noqa -- importing this adds readline behaviour to input()
+import tempfile
 import time
 from urllib.parse import unquote, urlparse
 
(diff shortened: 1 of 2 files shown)


evansd (approved):
Looks good

cohortextractor/presto_utils.py
```diff
@@ -38,23 +38,52 @@ def presto_connection_from_url(url):
...
+    pfx_path = env.get("PFX_PATH")
+
+    # unencrypted cert in env
+    if crt and key:
+        session.cert = (
+            write_to_temp_file(crt, prefix="presto_cert"),
+            write_to_temp_file(key, prefix="presto_key"),
+        )
+
+    # support encyrpted cert
```
sebbacon, line 69: *encrypted

PR #488, opened 2021-03-10
fix: Fix date arithmetic for with_vaccination_record in EMIS backend

inglesp:
This is a good illustration of the limits of the existing approach to building SQL queries!


sebbacon (approved):
> This is a good illustration of the limits of the existing approach to building SQL queries!

Isn't it just!

PR #501, opened 2021-03-18
fix: update Dockerfile to use 20.04 ppa for mssql

bloodearnest:


evansd (approved):
:+1: Hoped it might be that simple!

PR #502, opened 2021-03-19
Update docstrings in patients.py

HelenCEBM:
APCS docstrings;
improved formatting of docstrings throughout to improve readability when displayed in documentation, particularly for `returning` options.

Notes:
- the household variable is not currently rendered properly in docs - I removed a tab but not sure if this will help

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -747,7 +747,7 @@ def address_as_of(
 
     Args:
         date: date of interest as a string with the format `YYYY-MM-DD`. Filters results to the given date.
-        returning: a str defining the type of data to be returned. options include `index_of_multiple_deprivation`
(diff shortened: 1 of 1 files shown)


evansd:
I've discovered we use markdown bullet points if we add a blank line before the first bullet, which makes things a lot more readable. e.g.

127 0 0 1_8000_study-def-variables_

instead of

docs opensafely org_en_latest_study-def-variables_

I've gone through and applied this to all `returning` option documentation, but we could probably make use of this elsewhere as well.

/cc @wjchulme @CarolineMorton for interest

evansd (approved):

HelenCEBM:
Great thanks @evansd ! Can we do this for `Returns` as well, where this is a list?

evansd:
Yes, we should do that too. Just tackling it in small chunks for now!

PR #506, opened 2021-03-24
Remove duplicate registration IDs before running a study in EMIS

inglesp:


iaindillingham:
More of me saying to you what you've said to me. Thank you for your patience! It's good for comprehension, although it can sound like a conversation with a psychotherapist. So...

We're going to create `patient_all_orgs_v2` on each run, rather than creating `patient_all_orgs_v2` after a build. This means that we don't have to remember to create `patient_all_orgs_v2` after a build; or remember but create it during a run. However, maybe I've got that wrong: maybe we're going to create `patient_no_duplicates` rather than `patient_all_orgs_v2`?

Either way, first thought: can runs run in parallel?

If I understand your comment correctly, then SQLAlchemy/SQL Server won't allow you to create a FK to a column that isn't a PK. And because it's a PK, it can't have duplicate values. This surprised me! Do you point the finger of blame at SQLAlchemy or at SQL Server?

cohortextractor/emis_backend.py
```diff
@@ -20,30 +20,7 @@
...
-#     FROM patient_all_orgs_v2
-#     WHERE registration_id NOT IN (
-#         SELECT registration_id
-#         FROM patient_all_orgs_v2
-#         GROUP BY registration_id
-#         HAVING COUNT(*) > 1
-#     )
-# )
-PATIENT_TABLE = "patient_no_duplicates"
+PATIENT_TABLE = "patient_all_orgs_v2"
```
iaindillingham, line 23: I'm a bit confused by this switch.

cohortextractor/emis_backend.py
```diff
@@ -193,7 +173,19 @@ def get_queries(self, covariate_definitions):
...
-        all_queries = []
+        patient_no_duplicates_query = f"""
+        CREATE TABLE IF NOT EXISTS {self.patient_no_duplicates_table} AS (
+            SELECT *
+            FROM {PATIENT_TABLE}
+            WHERE registration_id NOT IN (
+                SELECT registration_id
+                FROM {PATIENT_TABLE}
+                GROUP BY registration_id
+                HAVING COUNT(*) > 1
```
iaindillingham, line 184: Why the `HAVING` in this sub-query? I've not used SQL Server; does it not have a `SELECT DISTINCT`, for example?

inglesp:
> More of me saying to you what you've said to me. Thank you for your patience! It's good for comprehension, although it can sound like a conversation with a psychotherapist. So...
> 
> We're going to create `patient_all_orgs_v2` on each run, rather than creating `patient_all_orgs_v2` after a build. This means that we don't have to remember to create `patient_all_orgs_v2` after a build; or remember but create it during a run. However, maybe I've got that wrong: maybe we're going to create `patient_no_duplicates` rather than `patient_all_orgs_v2`?

Not quite -- `patient_all_orgs_v2` is the name of the patient table in the database that EMIS build for us.  We were previously creating our own `patient_no_duplicates` table, using the SQL in the deleted comment, which contains all rows from `patient_all_orgs_v2` except those with duplicate `registration_id`s.

Now, at the start of each run, we create a temporary table called eg `_20200325_1254_abcd_patient`, which contains all rows from `patient_all_orgs_v2` except those with duplicate `registration_id`s.

This means that we don't have to remember to recreate `patient_no_duplicates` after a database refresh.

> Either way, first thought: can runs run in parallel?

Yes they can.  (In theory!  In practice, doing so might cause Presto to grind to a halt.)  But we work around this, by adding a prefix to the name of each temporary table we create.

> If I understand your comment correctly, then SQLAlchemy/SQL Server won't allow you to create a FK to a column that isn't a PK. And because it's a PK, it can't have duplicate values. This surprised me! Do you point the finger of blame at SQLAlchemy or at SQL Server?

I think both SQLAlchemy and SQL Server are being reasonable!   From a data modelling perspective, a foreign key to a column with duplicates doesn't really make sense.

cohortextractor/emis_backend.py
inglesp: We're interested in the patients that do not have duplicates.  For instance, if we have two entries for the same `registration_id` corresponding to a 23yo man from Gillingham (Kent) and a 93yo woman from Gillingham (Dorset), we want to ignore both, since there's no way of working out whether an observation (eg a diagnosis) belongs to one of the other.

Does that make sense?

cohortextractor/emis_backend.py
inglesp: Have I explained this in https://github.com/opensafely-core/cohort-extractor/pull/506#issuecomment-806698656?

cohortextractor/emis_backend.py
sebbacon: I knew this, but in reading your (excellent) explanation I realised it'd be very good to include a version of this explanation as a comment.

inglesp: Good point.  Have added a comment.

iaindillingham: That makes sense and is an excellent explanation. (👍🏻 for using the place I grew up and the place my family used to go on holiday every year, too!) Thank you for adding the comment as my first thought -- as you will have guessed -- was "Don't we just want a list of registration IDs?"

cohortextractor/emis_backend.py
iaindillingham: You have. Thank you 🙂

sebbacon (approved):
This makes sense from code-reading.

Is this code touched by tests? If not, is there a cheap way to smoketest it, to avoid having to wait a long time to see it break in production?

inglesp:
> Is this code touched by tests?

Yes -- pretty much every test runs through this code path.

evansd (commented):
This makes sense and looks like a sensible move. It does mean we redoing the de-duping work on every extract we run, but I suppose it's not a significant amount of work in the context of everything else we're doing.

PR #509, opened 2021-03-29
fix: Output consistent patient IDs

evansd:
Previously we generated synthetic sequential IDs for each extract but
this made it impossible to match patients across extracts which we
sometimes need to do. We now use just part of the original hash:

> EMIS patient IDs are 128 byte strings which significantly bloat our output
> files. However, because they are actually hex-encoded SHA-512 hashes we can
> get away with throwing away all but the first 63 bits and storing them as
> int64s (63 bits so we can safely store them as signed int64s without
> worrying about negative patient IDs, which might upset expectations
> elsewhere). 63 bits still gives us plenty of entropy as sqrt(2^63) is about
> 2 billion so well above the threshold where we'd need to worry about
> collisions with a population of a few tens of millions.

Note that this will produce larger CSVs than previously as patient IDs
will be up to 19 digits long whereas previously they would be at
most 8 digits. But this is still better than the 128 characters they
were initially. And the difference will disappear when using binary
output formats.

### DIFF
--- a/cohortextractor/emis_backend.py
+++ b/cohortextractor/emis_backend.py
@@ -68,11 +68,18 @@ class EMISBackend:
             nonlocal total_rows
             headers = [x[0] for x in result.description]
             id_column_index = headers.index("patient_id")
+            # In production "patient_id" is a string (hex-encoded hash) but in
+            # test it's an int so we need to handle that correctly
+            if result.description[id_column_index][1] == "integer":
+                truncate_patient_id = lambda x: x  # noqa
+
             yield headers
+
             for row in result:
-                unique_ids.add(row[id_column_index])
-                # Replace long hex ID with sequential int for space reasons
-                row[id_column_index] = total_rows
+                # Reduce long EMIS patient IDs (see docstring)
+                patient_id = truncate_patient_id(row[id_column_index])
+                row[id_column_index] = patient_id
+                unique_ids.add(patient_id)
                 total_rows += 1
                 if total_rows % 1000000 == 0:
                     logger.info(f"Downloaded {total_rows} results")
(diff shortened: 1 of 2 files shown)


inglesp (approved):

bloodearnest:
Two thoughts:

1) Is it worth checking the length of the input hex in truncate_id? E.g.s uppose EMIS change to 64 character ids, would the sqrt rule of thumb maths then be off? Maybe only truncate if >= 16 characters?

2) How would this affect the possibility of linking patients in EMIS and TPP backends? Is that even a thing? I know they are supposed to share a hash?

sebbacon:
> How would this affect the possibility of linking patients in EMIS and TPP backends? Is that even a thing? I know they are supposed to share a hash?

This is not a thing, they're not supposed to share a hash normally. There's an exception for a one-off duplicates check which has been an epic of many months, but that's different from our business-as-usual

evansd:
On 1, I think that would be a big enough change that it would have knock-on implications elsewhere and wouldn't be done silently. Remember that these are hashes (`secret_hasher(a."registration_id", '##INTERNAL_SALT##', 'sha512',false) as "registration_id"`) and so regardless of what happens with the internal registration ID the hash will stay the same length.

bloodearnest:
> On 1, I think that would be a big enough change that it would have knock-on implications elsewhere and wouldn't be done silently. Remember that these are hashes (`secret_hasher(a."registration_id", '##INTERNAL_SALT##', 'sha512',false) as "registration_id"`) and so regardless of what happens with the internal registration ID the hash will stay the same length.

Right, my concern was if the changed the hash algo to sha256 or lower, possibly to reduce CPU on import. That said, as long is it's more than 16 hex characters, it's all good.

PR #516, opened 2021-03-30
fix: Stop writing output to temporary filename first

evansd:
In the old days cohortextractor was run manually and it was important
that we didn't create files that looked like they were successful
outputs when they in fact had errors. We did this by writing outputs to
temporary filenames and then renaming them on success or, on error,
leaving them under their temporary names for debugging purposes.

Now that the job-runner infrastructure explicitly tracks exit codes we
don't have the same problem as it's impossible to accidentally use the
output of a failed job in another action. So using temporary files
brings unnecessary complexity. In fact, it's now actively harmful
because in the case that a cohort extraction fails the file produced
doesn't have the expected name and so doesn't match the output pattern
and is never copied out of the container. This means partially completed
output files aren't available for debugging, which is annoying.

### DIFF
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -1,7 +1,6 @@
 import datetime
 import enum
 import hashlib
-import os
 import re
 import uuid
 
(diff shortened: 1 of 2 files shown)


sebbacon (approved):
Mmmm, smells *good*!

PR #519, opened 2021-04-01
fix: Extend "prefix matching" to primary diagnosis field

### LINKED ISSUE #512: 3 character ICD-10 codes have `X` appended to them in SUS tables
The majority of ICD-10 codes are 4 characters, e.g. `I676`, but some are 3 characters, e.g. `G08`. In the SUS data, it appears that there are no recorded  3 char codes, but instead, they all have `X` appended to them: `G08X`

This appears to be true at least for the `APCS` table, but I haven't checked other SUS tables.

*However* we can't just append `X` to all 3 char ICD-10 codelists, because in the `ONS_Deaths` table has 3 char codes as you'd expect.

evansd:
When querying the Diagnoses field in the APCS data we use prefix
matching so that searching for the code `J12` will also return the code
`J120`, `J121` etc.

I initially thought this was a bug but there turns out to be a test for
precisely this behaviour. It also makes clinical sense given the
hierarchical nature of ICD-10 codes.

But we weren't doing this for the Primary Diagnosis field which is a
confusing inconsistency. It's particularly problematic with the APCS data
which seems to represent three characters codes by padding them with an
X to make them four character. As the resulting codes were not valid
ICD-10 codes there was no way to represent them in the codelist builder
and therefore. Using prefix matching sidesteps this problem.

This is possibly a slightly controversial approach; see in particular
the discussion at:
https://github.com/opensafely-core/cohort-extractor/issues/512#issuecomment-812027119

Closes #512

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -1904,8 +1904,12 @@ def admitted_to_hospital(
         date_format: a string detailing the format of the dates to be returned. It can be `YYYY-MM-DD`,
             `YYYY-MM` or `YYYY` and wherever possible the least disclosive data should be returned. i.e returning
             only year is less disclosive than a date with day, month and year.
-        with_these_diagnoses: icd10 codes to match against any diagnosis
-        with_these_primary_diagnoses: icd10 codes to match against the primary diagnosis
+        with_these_diagnoses: icd10 codes to match against any diagnosis (note
+            this uses **prefix** matching so a code like `J12` will match
+            `J120`, `J121` etc.)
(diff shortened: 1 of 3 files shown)


inglesp (approved):

sebbacon (approved):
I agree that this is the right approach for now; I also think if we were revisiting all our APIs I'd want prefix matching to be via a wildcard expression or some other explicit form. Hopefully that'll be an obvious choice when refactoring because we'll be reading through all the function signatures and thinking afresh.

PR #526, opened 2021-04-13
fix: Support patients_date_deregistered_from_all_supported_practices …

inglesp:
…in EMIS backend

### DIFF
--- a/cohortextractor/emis_backend.py
+++ b/cohortextractor/emis_backend.py
@@ -1091,6 +1091,20 @@ class EMISBackend:
               {self.patient_no_duplicates_table}
             """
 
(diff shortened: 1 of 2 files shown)


cohortextractor/emis_backend.py
```diff
@@ -1091,6 +1091,22 @@ def patients_registered_practice_as_of(self, date, returning=None):
               {self.patient_no_duplicates_table}
             """
 
+    def patients_date_deregistered_from_all_supported_practices(self, between):
+        if between is None:
+            between = (None, None)
```
evansd, line 1096: This is redundant I think, as `get_date_condition` handles this.

evansd (approved):
This looks correct to me

PR #537, opened 2021-04-22
Treat NULL observations dates as 1900-01-01

### LINKED ISSUE #522: Treat null dates as 1900-01-01 in EMIS
#### Problem
Currently, clinical events with no known "applicable" date are NULL in EMIS (this is the date we use in cohortextractor which represents the date the information applies to, rather than when it was entered). In TPP unknown dates default to `1900-01-01`. This means that these events are treated differently in cohort-extractor: (a) events with null dates are excluded when an `on-or-before` date is entered, and (b) if no date limits are entered, a result is returned for all `returning` options other than `date`, which returns a null, not indicating that the patient had actually matched the given criteria.  

Further discussion in [this](https://github.com/opensafely/emis-qa/issues/4) issue. 

#### Solution
Convert null dates to `1900-01-01` (or make sure cohortextractor treats them as such).

evansd:
Sometimes a clinician will want to record that an event has happened at some
point in the past without knowing the specific date. In TPP such events are
given a date of 1900-01-01 and in EMIS they are given a date of NULL.
However, we already use NULL as a value to indicate that a patient had no
matching event at all, so we need another value to indicate "they had a
matching event but at an indeterminate time in the past". For consistency's
sake we use the same date as in TPP.

Closes #522

### DIFF
--- a/cohortextractor/emis_backend.py
+++ b/cohortextractor/emis_backend.py
@@ -863,8 +863,12 @@ class EMISBackend:
         ignore_missing_values=False,
     ):
         codelist_table, codelist_queries = self.create_codelist_table(codelist)
+        # Using COALESCE like this has the potential to produce very
+        # inefficient queries by scuppering use of indices. However, my
+        # understanding is there are no indicies anyway in the current Presto
+        # setup so this shouldn't actually make much difference
         date_condition, date_joins = self.get_date_condition(
-            from_table, "effective_date", between
+            from_table, f"COALESCE(effective_date, {quote(EARLIEST_DATE)})", between
         )
         ignored_day_condition, extra_queries = self._these_codes_occur_on_same_day(
             from_table, ignore_days_where_these_codes_occur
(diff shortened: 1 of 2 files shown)


inglesp (commented):
It's not obvious to me why we want to return $EARLIEST_DATE for a patient with two matching events, where one event has a null date.  The alternative is to return the earliest non-null date -- is that something you considered?

evansd:
I don't think there is an obvious right answer in this case. If you know that a patient had an X event in 2001 and also at some indeterminate time in the past then, plausibly, the answer to the question "when did this patient first have an X event?" is " some indeterminate time in the past". And given this ambiguity I think the best course of action is to maintain consistency between the two backends.

sebbacon:
Just a drive-by comment to say - we need to remember to document this super-clearly somewhere. Not sure how / where! There is a "dates" section in the docs...

inglesp (approved):
Consistency's good.

PR #541, opened 2021-04-23
fix: Use strings not categoricals when saving as Stata

### LINKED ISSUE #534: Chaotic encoding of categorical variables in `.dta` output format
In contrast with dates (as described in #533) categorical columns that are created by e.g. `patients.categorised_as` _are_ preserved in the `.dta` output format. However, because of the slightly odd way that Stata encodes categorical variables, this can lead to some unpredictable behaviour.

Categorical variables in Stata are encoded as an `int`, with a label associated with each value. Unless manually specified, when a variable is [encoded](https://www.stata.com/manuals/dencode.pdf), values are assigned a zero indexed integer in alphabetical order. For example the categories `0-44, 45-69, 70+` would be encoded as `0, 1, 2`. This appears to be the behaviour that occurs when using `df.to_stata()` here.https://github.com/opensafely-core/cohort-extractor/blob/a5ec79afc4c01b1541197a04e0852c4268349f35/cohortextractor/pandas_utils.py#L11

This would work fine sometimes, but could cause unpredictable encoding. For example, if there were no people in the `0-44` category, then `45-69, 70+` categories would be assigned `0, 1` instead of `1, 2`.

Worse than this is categories that are returned as numbers, for example: the way that ethnicity has been defined in most cases so far, e.g. [here](https://github.com/opensafely/long-covid/blob/19609d7caaaf9303ea191b5492171597dde9d077/analysis/common_variables.py#L89-L98). This returns a categorical variable with the numbers 1-5, which `df.to_stata()` dutifully encodes as `0, 1, 2, 3, 4` and labels as `1, 2, 3, 4, 5`.
______
I'm not sure what the solution is here, as I don't think `df.to_stata()` has a way of specifying how categorical variables are encoded.

The solution that would at least restore sanity would be to change categorical variables to strings/ints as appropriate and then let people categorise in the downstream Stata. Obviously we then lose the memory/speed advantage.

evansd:
Stata's categorical support isn't quite up the job, unfortunately.

Closes #534

### DIFF
--- a/cohortextractor/pandas_utils.py
+++ b/cohortextractor/pandas_utils.py
@@ -8,6 +8,15 @@ def dataframe_to_file(df, filename):
     elif filename.endswith(".feather"):
         df.to_feather(filename)
     elif filename.endswith(".dta") or filename.endswith(".dta.gz"):
+        # Make a shallow copy of the dataframe so we can replace certain
+        # columns without duplicating the entire thing in memory
+        df = df.copy(deep=False)
+        # Replace categorical columns with arrays of strings as Stata's
+        # categorical support isn't that great. See:
+        # https://github.com/opensafely-core/cohort-extractor/issues/534
+        for column, dtype in list(df.dtypes.items()):
+            if isinstance(dtype, pandas.Categorical):
+                df[column] = df[column].__array__()
         # Encode dates as proper Stata dates (we use "td" as we only need day
         # resolution)
         convert_dates = {


ghickman (approved):

cohortextractor/pandas_utils.py
```diff
@@ -8,6 +8,15 @@ def dataframe_to_file(df, filename):
...
     elif filename.endswith(".dta") or filename.endswith(".dta.gz"):
+        # Make a shallow copy of the dataframe so we can replace certain
+        # columns without duplicating the entire thing in memory
+        df = df.copy(deep=False)
+        # Replace categorical columns with arrays of strings as Stata's
+        # categorical support isn't that great. See:
+        # https://github.com/opensafely-core/cohort-extractor/issues/534
+        for column, dtype in list(df.dtypes.items()):
+            if isinstance(dtype, pandas.Categorical):
+                df[column] = df[column].__array__()
```
sebbacon, line 19: Apols I've not time to check myself but isn't `df[column].astype(str)` what we want here?

evansd: Could well be. I don't really know Pandas. This seemed to do what I wanted though.

sebbacon: Now worries, it's just the ergonomics of using dunder methods making me a bit sad, but /shrug

PR #547, opened 2021-04-29
docs: Clarify behaviour of carehome function

CarolineMorton:
Where a patient does not have any matching address it will produce an
empty string as default.

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -836,7 +836,9 @@ def care_home_status_as_of(
 
     Example:
 
-        This creates a variable called `care_home_type` which contains a 2 letter string which represents a type of care home environment.
+        This creates a variable called `care_home_type` which contains a 2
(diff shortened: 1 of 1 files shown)


evansd (approved):
Good move

evansd:
@CarolineMorton Ah, you'll need to change the commit title to have to prefix `doc:`

It might also be worth adding something in the commit body explaining why we're making the change (doesn't have to be loads of detail).

You should be able to do that with
```
git commit --amend
gt push -f
```
but give me a shout if you'd like me to do it.

CarolineMorton:
@evansd What am I doing wrong here?

evansd:
@CarolineMorton I don't know, sorry! Git is a mysterious and cantankerous beast at times :smile:  I'll sort it.

evansd:
Sorry, the prefix was `docs` :disappointed: My fault!

PR #550, opened 2021-05-06
Fix ICU queries when used with dynamic dates

evansd:
The old SQL we used was weird in that it applied the date range filter
condition to the aggregated (min/max) date using a HAVING clause, rather than
filtering with a WHILE and then aggregating. This was introduced in
5ac9039, I suspect just because this was the most straightforward change
to make at the time not because there's any special reason to do the
query this way. This PR brings the SQL in line with the patterns we use
elsewhere and makes it work with dynamic dates.

We also update the equivalent query in the EMIS backend to fix this bug, support
the additional options we've implemented in TPP and to match the behavioural
change we introduced in #284.

### DIFF
--- a/cohortextractor/emis_backend.py
+++ b/cohortextractor/emis_backend.py
@@ -1337,43 +1337,55 @@ class EMISBackend:
         find_last_match_in_period=None,
         returning="binary_flag",
     ):
-        if find_first_match_in_period:
-            date_aggregate = "MIN"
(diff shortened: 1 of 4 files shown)


inglesp (changes requested):
Can you make the changes to the EMIS backend too?

inglesp (approved):
Thanks!

PR #562, opened 2021-05-28
feat: Support variant data in SGSS

evansd:
Full list of possible variant values at present is below, but this is obviously subject to change.

Screenshot from 2021-05-28 10-26-19


sebbacon (approved):

Bengoldacre:
blimey that was quick

PR #565, opened 2021-06-08
fix: Don't leak internal 9999-21-31 deregistration date

### LINKED ISSUE #500: patients.date_deregistered_from_all_supported_practices returns `9999-12-31` instead of `null`
The default value in the TPP `RegistrationHistory` table for the `EndDate` is `9999-12-31`, and currently the backend for `patients_date_deregistered_from_all_supported_practices` runs this query: https://github.com/opensafely-core/cohort-extractor/blob/19df1d811259971dca0c6c59f2c79747e389e02b/cohortextractor/tpp_backend.py#L1220-L1230 which returns this `9999-12-31` date where a patient has not deregistered. It should instead return a `null` value.

evansd:
Currently active registrations are recorded as having an end date of
9999-12-31. This is an internal detail we shouldn't leak: we should
instead return a deregistration date of NULL for currently registered
patients. Aside from being inelegant, this date can't be represented as
a Pandas timestamp and so it causes binary exports to fail.

We already attempted to do the right thing here, but the code was buggy
and the tests inadequate.

Closes #500

### DIFF
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -1335,7 +1335,9 @@ class TPPBackend:
             max_date = "3000-01-01"
         if min_date is None:
             min_date = "1900-01-01"
-        date_condition, date_joins = self.get_date_condition("t", "t.end_date", between)
+        date_condition, date_joins = self.get_date_condition(
+            "t", "t.end_date", (min_date, max_date)
+        )
         return f"""
         SELECT
           t.Patient_id,
--- a/tests/test_tpp_backend.py
+++ b/tests/test_tpp_backend.py
@@ -2741,8 +2741,16 @@ def test_patients_date_deregistered_from_all_supported_practices():
             on_or_before="2018-02-01",
             date_format="YYYY-MM",
         ),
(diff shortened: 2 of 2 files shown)


cohortextractor/tpp_backend.py
```diff
@@ -1335,7 +1335,9 @@ def patients_date_deregistered_from_all_supported_practices(self, between):
             max_date = "3000-01-01"
         if min_date is None:
             min_date = "1900-01-01"
-        date_condition, date_joins = self.get_date_condition("t", "t.end_date", between)
```
inglesp, line 1338: Should a linter have picked up that `max_date` and `min_date` were never used?

inglesp (approved):

cohortextractor/tpp_backend.py
evansd: Sadly not, because they appear in conditionals. So this is a pyflakes error:
```py
def foo():
    bar = 1
```
But this isn't:
```py
def foo():
    bar = 1
    if bar:
        pass
```

PR #580, opened 2021-06-11
Add skeletal Databricks backend

inglesp:


evansd (approved):
Excellent.

I need to try to get my head around the execution model Spark uses here. Is running:
```py
dataframes[name] = self.session.sql(sql)
```
Sort of equivalent to SQL Server's
```sql
SELECT * INTO {name} FROM ({sql}) t
```
In that it creates a temporary object on the server? Or is it pulling down data locally and then doing the final join there?

inglesp:
> Excellent.
> 
> I need to try to get my head around the execution model Spark uses here. Is running:
> 
> ```python
> dataframes[name] = self.session.sql(sql)
> ```
> 
> Sort of equivalent to SQL Server's
> 
> ```sql
> SELECT * INTO {name} FROM ({sql}) t
> ```
> 
> In that it creates a temporary object on the server? Or is it pulling down data locally and then doing the final join there?

`session.sql(sql)` returns a `pyspark.sql.dataframe.DataFrame`, which is unevaluated.  We combine a bunch of `DataFrame`s in `.execute()`, but it's not until the last line (`result.collect()`) that any evaluation actually happens.  The final join happens "in Spark", whatever that means, so we're only locally pulling down the final results.

Does that make sense?

evansd:
Yes, perfectly! Thanks

PR #588, opened 2021-06-17
fix: Presto and MSSQL functions were the wrong way round

evansd:
This wasn't caught previously as the tests didn't exercise this code and
no one had used it in production.


inglesp (approved):
Thanks.  Should we be aiming for 100% test coverage here?

evansd:
I think just having a coverage report would be an extremely useful first step.

PR #592, opened 2021-06-17
Pin python wheel in Github Actions

madwort:


madwort:
I don't remember why I wanted to do this, but it's been 2 years without it so I guess it's ok.

PR #593, opened 2021-06-21
chore: move to use new base-action image

bloodearnest:


evansd (commented):
Isn't this effectively a no-op, as all base-action does over base-docker is provide a default entrypoint script and cohortextractor defines its own entrypoint? Or will there be further advantages down the line?

evansd:
Ah test failure is nothing to do with the PR and is a thing I've been meaning to change for a while:
```
    def test_docker_version():
        with open("docker-compose.yml") as f:
            our_version = yaml.safe_load(f)["services"]["presto"]["image"].split(":")[1]
    
        rsp = requests.get("https://explorerplus.emishealthinsights.co.uk/v1/info")
        their_version = rsp.json()["nodeVersion"]["version"]
    
>       assert our_version == their_version
E       AssertionError: assert '350-e.3' == '350-e.8'
E         - 350-e.8
E         ?       ^
E         + 350-e.3
E         ?       ^
```
We check the version of Presto we use in tests matches the live version. But I think this should be a scheduled check rather than part of the test suite.

bloodearnest:
> Isn't this effectively a no-op, as all base-action does over base-docker is provide a default entrypoint script and cohortextractor defines its own entrypoint? Or will there be further advantages down the line?

It's a no-op currently, but in the middle of changes for cohortextractor to reuse the entrypoint.sh in base-action, so it won't be forever.

Also, if I don't do it now, I'll forget and then spend far too long figuring why something isn't working the future.

bloodearnest:
> Ah test failure is nothing to do with the PR and is a thing I've been meaning to change for a while:
> 
> ```
>     def test_docker_version():
>         with open("docker-compose.yml") as f:
>             our_version = yaml.safe_load(f)["services"]["presto"]["image"].split(":")[1]
>     
>         rsp = requests.get("https://explorerplus.emishealthinsights.co.uk/v1/info")
>         their_version = rsp.json()["nodeVersion"]["version"]
>     
> >       assert our_version == their_version
> E       AssertionError: assert '350-e.3' == '350-e.8'
> E         - 350-e.8
> E         ?       ^
> E         + 350-e.3
> E         ?       ^
> ```
> 
> We check the version of Presto we use in tests matches the live version. But I think this should be a scheduled check rather than part of the test suite.

Yep, and doing a separate PR to fix

bloodearnest:
Gonna do this a bit differently I think

evansd:
> It's a no-op currently, but in the middle of changes for cohortextractor to reuse the entrypoint.sh in base-action, so it won't be forever.
> 
> Also, if I don't do it now, I'll forget and then spend far too long figuring why something isn't working the future.

:+1: Makes sense. Just wanted to check

PR #596, opened 2021-06-29
Fix bug with "group by population"

evansd:
This fixes a relatively simple, though annoying to track down, bug in the measures framework when using the "population" column as both a numerator (to get per-capita values) and as a group_by column (to treat the entire population as a single group). In this case, the group_by operation overrides the sum, so that the population column ends up with a value of 1, rather than the sum of the population. There was a measure exercising this behaviour in the test fixture, but nothing was asserted about its output.

This PR correctly handles the "population" case, and raises an error for any other cases of overlap between numerator/denominator and group_by columns.

### DIFF
--- a/tests/test_smoketest.py
+++ b/tests/test_smoketest.py
@@ -14,12 +14,13 @@ import cohortextractor
 
 @pytest.mark.parametrize("format", ["csv", "csv.gz", "feather", "dta", "dta.gz"])
 def test_smoketest(tmp_path, format):
+    population_size = 100
     _cohortextractor(
         study="test_smoketest",
         args=[
             "generate_cohort",
             "--expectations-population",
-            "100",
+            str(population_size),
             "--index-date-range",
             "2020-01-01 to 2020-04-01 by month",
             "--output-dir",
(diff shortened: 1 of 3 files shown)


inglesp (approved):
nice clear explanation

tests/test_smoketest.py
```diff
@@ -63,16 +66,30 @@ def test_smoketest(tmp_path, format):
...
+
     # liver_disease
     with open(tmp_path / "measure_liver_disease.csv") as f:
         contents = list(csv.reader(f))
+    assert contents[0] == [
+        "has_chronic_liver_disease",
+        "population",
+        "value",
+        "date",
+    ]
```
rebkwok, line 78: Could there be an assert that makes it clear how this one differs from the one-group version?  Does it differ, if it has no groupby?

rebkwok (approved):

tests/test_smoketest.py
evansd: Ah yeah, good point. The tests are very much not great here. The difference is that in the "no group" version every patient is on their own line so the assert would be something like "number of rows = population X number of dates" and "population = 1 for every row".

This all started as a very basic smoke test (as the name implies) that the main entrypoint was functioning OK. Then when the measures stuff was added it was a quick and dirty way of adding some kind of testing for it then was always intended to be replaced with "proper testing" at some point and then never was :)

PR #598, opened 2021-06-30
More databricks work

inglesp:


cohortextractor/databricks_backend.py
```diff
@@ -3,11 +3,20 @@
...
     def __init__(self, database_url, covariate_definitions, temporary_database=None):
-        self.next_temp_table_id = 1
-        self.session = self.get_spark_session(database_url)
+        self.next_temp_table_id = 0
+        self.session = self.create_spark_session()
         self.dataframes = self.build_dataframes(covariate_definitions)
 
+    def create_spark_session(self):
+        conf = SparkConf().set(
+            "spark.driver.extraClassPath", "sqljdbc_9.2/mssql-jdbc-9.2.1.jre11.jar"
```
evansd, line 18: Is this relative to the cwd? I wonder if you need to do something with `__file__` here if this is to work when cohortextractor is invoked inside the `/workspace` directory.

tests/databricks_backend_setup.py
```diff
@@ -31,46 +31,25 @@ def make_database():
...
+        "CodedEvent_ID": Column(Integer, primary_key=True),
+        "CTV3Code": Column(String(collation="Latin1_General_BIN")),
+        "NumericValue": Column(Float),
+        "ConsultationDate": Column(DateTime),
+        "SnomedConceptId": Column(String),
+    }
+    return type(f"CodedEvent{year}", (Base,), columns)
+
+
+CodedEvent2018 = build_coded_event_table(2018)
```
evansd, line 50: Ouch, this is going to make life tricky, isn't it?

evansd (approved):

cohortextractor/databricks_backend.py
inglesp: Quite probably.  Will fix.

tests/databricks_backend_setup.py
inglesp: yep

sebbacon (commented):
Re the databricks connection string, this is from an email they sent a while ago

>>    We need to work out how to get a test infrastructure set up locally. My understanding matches Simon's: that Databricks is simply querying an Apache Spark over ODBC. We need to check this assumption, and get the Spark version, in order to set up the test infrastructure. Who can help us with this query? Can probably be resolved by email
>
> Yes, the assumption is right. Also, the Databricks runtime version our cluster is currently using is 6.6 (includes Apache Spark 2.4.5, Scala 2.11)
>
> The connection string we will be providing looks something like this:  `jdbc:spark://<host>:<port>/default;transportMode=http;ssl=1;httpPath=sql/protocolv1/o/0/<cluster-id>;AuthMech=3;UID=<uid>;PWD=<personal-access-token>`
> 
> I've also tried connection to Databricks before using the "databricks-dbapi" python library, which is a thin wrapper around pyhive and pyodbc for creating a DBAPI connection to Databricks (https://pypi.org/project/databricks-dbapi/). You might want to check if it works with Apache Spark for your local setup.
>
>
>>   Emmanuel and co were going to explore setting up a temporary develop environment we'd be able to access, so that we can test our Databricks backend db code against a real db connection (but dummy data), and iterate on it faster when we get to an initial integration stage, but we're not at that point yet.
>
> We've managed to setup a development environment which has got access to the Databricks backend for any test that we may want to do.

PR #602, opened 2021-07-02
Support pre-generated dummy data

inglesp:
Users can now provide cohort-extractor with a file of pre-generated dummy data with a new `--dummy-data-file` flag.

TODO: add documentation to docs.opensafely.org.


cohortextractor/validate_dummy_data.py
```diff
@@ -0,0 +1,114 @@
...
+            except ValueError:
+                raise DummyDataValidationError(
+                    f"Invalid value `{value}` for {col_name} in row {ix + 2}"
+                )
+
+
+def read_into_dataframe(path):
+    """Read data from path into a Pandas DataFrame."""
+
+    if path.suffixes in [[".csv"], [".csv", ".gz"]]:
```
ghickman, line 58: TIL about `.suffixes`, nice!

cohortextractor/validate_dummy_data.py
```diff
@@ -0,0 +1,114 @@
...
+
+    for col_name in df:
+        if col_name == "patient_id":
+            # patient_id is not present in covariate_definitions
+            continue
+
+        if col_name not in covariate_definitions:
+            raise DummyDataValidationError(
+                f"Unexpected column {col_name} in dummy data"
+            )
```
ghickman, line 23: Would it make sense here to find all the column names which don't exist in `covariate_definitions`, instead of failing on the first one?

ghickman (approved):
Looks great, one UX note but not sure if it's useful in this context.

cohortextractor/validate_dummy_data.py
```diff
@@ -0,0 +1,114 @@
...
+        return pd.read_csv(path)
+    elif path.suffixes in [[".dta"], [".dta", ".gz"]]:
+        return pd.read_stata(path)
+    elif path.suffix == ".feather":
+        return pd.read_feather(path)
+    else:
+        assert False, path
+
+
+def get_validator(col_name, query_args):
```
evansd, line 68: These validators are, I think, specific to CSV files and won't work with binary formats. At least the date ones certainly are and the `int`, `str`, `float` checks will test that columns can be cast to the right type but not that they're the right type initially.

Perhaps this function could take an extra argument which is the file format so it can return the appropriate validator. In practice I hope we'd only need two sets: one for CSV and another for all other binary formats.

cohortextractor/study_definition.py
```diff
@@ -71,6 +72,7 @@ def to_file(
             )
             dataframe_to_file(df, filename)
         elif dummy_data_file:
+            validate_dummy_data(self, dummy_data_file)
```
evansd, line 75: I think we need something to check that the dummy data file format matches the requested output format. That's almost as simple as checking that they have the same extension although the two level extensions (e.g. `.csv.gz`) make that a bit more tricky.

evansd (changes requested):
I think this is a really nice approach. Previously I was stuck on the idea that "data doesn't belong in git" and therefore that the code to generate the dummy data would need to be an action, but some sort of conditional action that only got run for certain backends and it all got very messy.

But of course the files are tiny in this case. And by just committing the dummy data to the repo we completely sidestep all of that, and people can use whatever tools they want to generate it with, which is great.

I've made a few comments on things which I think need addressing. Chiefly it's based around the fact that we now support multiple sorts of binary output and we need to validate them all.

cohortextractor/validate_dummy_data.py
```diff
@@ -0,0 +1,114 @@
...
+    """Read data from path into a Pandas DataFrame."""
+
+    if path.suffixes in [[".csv"], [".csv", ".gz"]]:
+        return pd.read_csv(path)
+    elif path.suffixes in [[".dta"], [".dta", ".gz"]]:
+        return pd.read_stata(path)
+    elif path.suffix == ".feather":
+        return pd.read_feather(path)
+    else:
+        assert False, path
```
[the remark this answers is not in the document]
inglesp: Good idea.

cohortextractor/validate_dummy_data.py
inglesp: Good idea, thanks.

cohortextractor/validate_dummy_data.py
```diff
@@ -0,0 +1,114 @@
...
+    df = read_into_dataframe(dummy_data_file)
+
+    covariate_definitions = study_definition.covariate_definitions
+
+    for col_name in df:
+        if col_name == "patient_id":
+            # patient_id is not present in covariate_definitions
+            continue
+
+        if col_name not in covariate_definitions:
```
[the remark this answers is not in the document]
inglesp: Good idea, thanks.

setup.py
```diff
@@ -18,7 +18,6 @@
         "opensafely-jobrunner>=2.0,<3.0",
```
evansd, line 18: This dependency can go as well now I think about it.

cohortextractor/study_definition.py
```diff
@@ -70,6 +73,15 @@ def to_file(
             df["patient_id"] = default_rng().choice(
                 (len(df) * 10), size=len(df), replace=False
             )
+            if not is_csv_filename(filename):
+                # Convert the data generated by make_df_from_expectations to the correct
+                # format for binary outputs.  This matches the approach taken in the
+                # backend .to_file methods.
+                buffer = StringIO()
```
evansd, line 81: Users sometimes hit memory issues when generating large dummy cohorts locally (particularly as Docker on mac/Win runs in a VM with a fixed memory allocation). This will approach will exacerbate that problem, but I think that probably the right thing to do is just to roll with it for now and address user issues if/when they arise.

evansd (approved):
Looks good to me!

PR #606, opened 2021-07-08
Make driver packages an optional dependency

evansd:
Although most usage of the cohortextractor is via the Docker image we built there are still two contexts in which we want to pip install it:

 1. When building the docs so we can pull the relevant docstrings and function signatures into the docs
 2. Inside the python-docker image so that researchers can import their own study definitions as part of their analysis pipeline.

In neither case do we need to actually talk to a real database and thus we don't need all the heavyweight driver dependencies.

This PR makes those optional dependencies (under the name `drivers`) and restores a test to ensure we don't accidentally import any of these when loading a study definition.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,10 +17,8 @@ setup(
         "lz4",
         "opensafely-jobrunner>=2.0,<3.0",
         "pandas",
-        "presto-python-client",
         "prettytable",
         "pyarrow",
-        "pyspark",
         "pyyaml",
         "requests",
         "requests-pkcs12",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -32,6 +30,26 @@
...
+            # you're on a platform other than Linux then you'll have to install
+            # cTDS yourself:
+            # https://zillow.github.io/ctds/install.html
+            #
+            # For more background on this see:
+            # https://github.com/opensafely/cohort-extractor/pull/286
+            "ctds@https://github.com/opensafely/ctds-binary/raw/9466f4bdb8eb70318256115c3bbb6b3ecc9351d0/dist/ctds-1.13.0-cp38-cp38-manylinux2014_x86_64.whl#egg=ctds;sys_platform=='linux'",
+            "pyodbc",
+        ]
+    },
```
bloodearnest, line 52: This is a really simple solution to a gnarly problem. Nice!

tests/test_study_definition.py
```diff
@@ -72,6 +75,46 @@ def test_errors_are_triggered_without_database_url(monkeypatch):
...
+    """
+    For now, we still have to support researchers importing their study
+    definitions in their analysis code to access details of their study
+    configuration (and avoid having to repeat this in two places). To enable
+    this we install a version of the cohortextractor inside the `python-docker`
+    image but without any of the driver packages (which are large and unused
+    and would bloat the image). But this means we need to make sure we don't
+    accidentally create a dependency on one of the driver packages when loading
+    a study definition. So this test ensures that we can import a basic study
+    defiition and generate some dummy data without ever importing a driver.
```
bloodearnest, line 88: s/defiition/definition

bloodearnest (approved):
Nice solution!

PR #612, opened 2021-07-14
feat: Add `symptomatic` as return option for SGSS

### LINKED ISSUE #581: Add `Symptomatic` column as a return option for SGSS
The `AllTests` versions of the SGSS tables contain a `Symptomatic` column, which [3 different users](https://ebmdatalab.slack.com/archives/C010SJ89SA3/p1623410097410800) have identified as being important for their studies.

It's not clear from the [schema](https://github.com/opensafely/database-notebooks/blob/master/notebooks/database-schema.ipynb) what exactly this contains, but it's quite likely that the raw column would be acceptable to return.

StevenMaude:
Fixes #581; see that issue for more context.

### DIFF
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -1838,7 +1838,6 @@ class TPPBackend:
                 "U": "",
                 "Y": "Y",
                 # From SGSS_AllTests_Negative table.
-                "NULL": "",
                 "false": "N",
                 "true": "Y",
             }
@@ -1847,6 +1846,13 @@ class TPPBackend:
                     f"WHEN {raw_column} = {quote(key)} THEN {quote(value)}"
                     for (key, value) in transforms.items()
                 ]
+                # As per #581, we only have nulls in the SGSS_AllTests_Negative table.
+                # However, it seems reasonable to apply this condition to both the
+                # SGSS_AllTests_Positive and SGSS_AllTests_Negative table.
+                # If nulls did appear in the SGSS_AllTests_Negative table, presumably
+                # we would want the same behaviour. Therefore, we can just add this
+                # extra clause unconditionally.
+                + [f"WHEN {raw_column} IS NULL THEN {quote('')}"]
             )
             column = f"CASE {case_clauses} ELSE {raw_column} END"
         else:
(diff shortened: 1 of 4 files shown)


sebbacon (commented):
I realise this is draft so you're probably already planning it, but just a reminder that docstrings will also need updating per Alex's observation _I think either way of doing it would be okay, so long as it's clearly documented_

sebbacon:
Another thing in passing - probably adding to `news.md` in the docs?

tests/test_tpp_backend.py
```diff
@@ -2324,6 +2349,9 @@ def test_patients_with_test_result_in_sgss():
         sgtf_result=["9", "1", "", ""],
         variant=["B.1.351", "VOC-20DEC-01", "", ""],
         variant_detection_method=["Private Lab Sequencing", "Reflex Assay", "", ""],
+        symptomatic_positive=["Y", "N", "", ""],
+        symptomatic_negative=["N", "", "", ""],
+        symptomatic_any=["N", "", "", ""],
```
madwort, line 2354: am I misunderstanding this, or should this be 
```
symptomatic_any=["N", "N", "", ""],
```

madwort: oh, no, I see - the `SGSS_AllTests_Negative` for this patient is first

inglesp (approved):
I suspect `+110 -14` doesn't do justice to the amount of work that's gone into this.  Chapeau.

cohortextractor/tpp_backend.py
```diff
@@ -1830,6 +1830,25 @@ def patients_with_test_result_in_sgss(
...
             column = "t2.variant_detection_method"
+        elif returning == "symptomatic":
+            raw_column = "t2.symptomatic"
+            transforms = {
+                # From SGSS_AllTests_Positive table.
+                "N": "N",
+                "U": "",
+                "Y": "Y",
+                # From SGSS_AllTests_Negative table.
+                "NULL": "",
```
inglesp, line 1841: Are you matching the string `"NULL"` or a literal SQL `NULL`?

inglesp (commented):
Actually, sorry, I've just spotted a potential issue with NULLs -- thoughts?

cohortextractor/tpp_backend.py
StevenMaude: Think I've fixed this now.

StevenMaude:
@inglesp: please take another look :eyes:

inglesp (approved):
That's better, thanks!

PR #617, opened 2021-07-21
fix: Remove validation that registration history date is recent

inglesp:
We have documented the limitation, and as far as I know, nobody has
ever noticed the warning.

We're doing this now, because validation breaks on non-literal date
expressions, which we need for a study.


evansd (approved):

inglesp:
Tests failing because of Presto version mismatch, which @StevenMaude is on.

Issue #242, opened 2020-07-17
Mismatch between "semantic-pull-request" check and "github-tag-action"


evansd:
By pre-pending "fix:" to [this PR](https://github.com/opensafely/cohort-extractor/pull/240) title, the PR passed the [semantic-pull-requests check](https://github.com/zeke/semantic-pull-requests).

However when merged it didn't then trigger a new release:
```
Analyzing commit: Merge pull request #240 from opensafely/sqlcmd-improvements
fix: Refactor sqlcmd wrapper for better error handling
The commit should not trigger a release
Analyzing commit: Be slightly more defensive against unexpected output
The commit should not trigger a release
Analyzing commit: Refactor sqlcmd wrapper for better error handling
Previously this code used the `-o output_file` option. This caused both
query output and warnings/errors to be mixed together in a single file,
which then required hairy and error prone code to try to separate back
out.

We now just let sqlcmd write to stdout, which we redirect to a file
ourselves, leaving error output written to stderr.

Closes #239
The commit should not trigger a release
Analyzing commit: Test that SQL runtime errors are handled correctly
The commit should not trigger a release
Analyzing commit: Compare contents of files directly in tests
This makes the test failure output more useful as it shows the diff
between the files.
The commit should not trigger a release
Analysis of 5 commits complete: no release
```

I think this is because the merge commit looks like this and doesn't contain "fix:" as its first characters:
```
Merge pull request #240 from opensafely/sqlcmd-improvements
fix: Refactor sqlcmd wrapper for better error handling
```

Ideally the PR check and the action would agree about what constitutes an appropriately annotated PR.

sebbacon:
Yes, `github-tag-action` considers only the first characters of commits.

I see that the conventional commits specification is that every commit *must* have a prefix, so I think the fix would be to make `semantic-pull-requests` observe this specification.  This has the advantage of consistency, and already being supported in the packages we're using.

However, I can also imagine there not being consensus on this so we should discuss!

I also note that the semantic-pull-requests is actually designed with [this CI tool](https://github.com/semantic-release/semantic-release) in mind, which is targetted at node apps but I've just noted can be configured for other setups.  Given the release tool I implemented already had a pretty egregious bug I had to fix, that might be worth considering....

PR #619, opened 2021-07-22
feat: Migrate from Presto to Trino in Starburst

### LINKED ISSUE #616: Migrate code to Starburst 356-e.5
In our tests, we check the Starburst version that EMIS run. They upgraded to Starburst 356-e.5. This has a breaking change of switching from PrestoSQL to Trino, though this is simply a [rebrand](https://trino.io/blog/2020/12/27/announcing-trino.html).

[From the information Starburst provide](https://docs.starburst.io/354-e/appendix/to-sep-354-plus.html) on this upgrade, it likely requires:
* switching Docker image
* replacing `presto-python-client` with `trino-python-client`
* some renames in our code

StevenMaude:
Fixes #616.

# Why is this change needed?

This is a necessary migration because:
* In our tests, we check the version of Starburst Enterprise that EMIS use and they have updated to a more current version.
* Current versions of Starburst Enterprise moved from Presto to Trino.

# What changes?

## Trino changes

Trino is a renamed fork of Presto. This does not change too much for us, but:

* The Docker image name has changed.
* There is a Trino Python client that replaces the Presto client. There do not seem to be any breaking API changes for us.
* The configuration location is now `/etc/starburst`.

## cohort-extractor changes

To make things work again, as well as for consistency and understanding in future, *almost* everything relating to Presto has been updated to refer to the new name/configuration. See below for what hasn't been updated.

### Breaking change

* `cohortextractor/presto_utils.py` is now `cohortextractor/trino_utils.py`, so you'll need to run `python -m cohortextractor.trino_utils DATABASE_URL` to use the Trino REPL.

### Non-breaking changes

* `presto://` still works, but you should replace this with `trino://`.
* `PRESTO_TLS_CERT` and `PRESTO_TLS_KEY` environment variables still work, but you should replace these with `TRINO_TLS_CERT` and `TRINO_TLS_KEY` *in future, when they are supported by opensafely-cli and job-runner*. The `TRINO` environment variables are used if both pairs of environment variables are provided.


StevenMaude:
The tests pass and I've renamed everything that I can find. There are a couple of points to note here.

## Renaming principles

I took the route of replacing "Presto" with "Trino" almost everywhere, in code, filenames and comments/docstrings. One place I *haven't* is the `presto/` directory which contains configuration files that override those in the Starburst Docker container. I renamed that to `starburst/` because the paths for the files actually and the `/starburst/etc/node.properties` file actually refers to `starburst` too.

I could change it to `trino/`; I've no strong opinion.

## Legacy environment variable

There are `PRESTO_TLS_CERT` and `PRESTO_TLS_KEY` environment variables. These are also used in job-runner and opensafely-cli. If we rename these here, they'll need renaming elsewhere and it may be that certain environments require them to be updated too.

On the other hand, it leads to ambiguity for anyone investigating this in future; there's an extra step involved in realising that Presto was the old name before Trino.

### Options

We could:
* Fix them up everywhere now.
* Live with these being legacy forever.
* Live with these being legacy for now and fix them up in later PRs.

evansd (changes requested):
Thanks @StevenMaude, this can't have been much fun but it needed doing and has been done with characteristic attention to detail!

I think that leaving the `starburst` stuff as it is is the right call.

As for the environment variables, I think the ideal approach would be to support _both_ the new and legacy names. That gives us the option to deploy, update other systems to use the new names, and then eventually remove the old ones at a leisurely pace without needing to coordinate updates to multiple systems.

We do have to continue to support the `presto://` URL scheme for now, which is why I've marked this as needing changes, but hopefully that's an easy change. Again, ideally we can do the same thing there and support both the old and new schemes simultaneously.

Other than that, all looks good. Thanks!

evansd:
Oh, just realised that although "chore" seems like a very accurate description here, it will need a prefix that generates a new version, either `fix:` or `feat:`.

starburst/etc/catalog/mssql.properties
```diff
@@ -1,4 +1,4 @@
 connector.name=sqlserver
-connection-url=jdbc:sqlserver://mssql:1433;databasename=Test_EMIS
+connection-url=jdbc:sqlserver://mssql:1433;database=Test_EMIS
```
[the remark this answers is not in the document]
StevenMaude: I'm not sure! It may not be required. This was added in the "let's try anything" stage of getting this to work.

Reference:

https://trino.io/docs/current/connector/sqlserver.html#configuration

StevenMaude:
It wasn't actually too bad. It was more the fact that the existing Starburst and Trino documentation is piecemeal. Also, one Starburst page I was referring to hasn't yet been correctly updated, which didn't help! [1]

@evansd: Allowing both schemes and environment variables seems a good move. We can open an issue/leave a comment that states to review this in e.g. 6-12 months time; thanks for thinking of that! :+1:

We'll also need to add those environment variables to job-runner and opensafely-cli. But that shouldn't block merging this, provided we don't actually use the new form of the environment variables yet. If we're the only consumers of this, then that's fine.

evansd:
Yes, we're the only consumers/users of this so we control both ends, as it were.

I have just remembered [this issue](https://github.com/opensafely-core/cohort-extractor/issues/242) which means that although changing the PR title passes the semantic versioning check, it won't actually create a new release unless the prefix is in one of the commit titles. Sorry!

evansd:
Oh sorry, I wasn't paying enough attention: you already had the right prefixes in the commit titles. I got fixated on the prefix in the PR title and just didn't notice. Ignore previous comment!

evansd:
Oh sorry, I wasn't paying enough attention: you already had the right prefixes in the commit titles. I got fixated on the prefix in the PR title and just didn't notice. Ignore previous comment!

StevenMaude:
Ah, I amended the first commit in this PR to read `feat:` so all the subsequent commits that were previously reviewed have changed IDs. However, I have not edited the commits apart from that. So the only new changes are now after bb277fc.

cohortextractor/trino_utils.py
```diff
@@ -93,9 +93,17 @@ def trino_connection_params_from_url(url):
...
     parts = parsed.path.strip("/").split("/")
-    if len(parts) != 2 or not all(parts) or parsed.scheme != "trino":
+    # presto:// is now legacy and replaced with trino://
+    # presto:// is included for backwards compatibilty only and can be
+    # removed in future.
+    if (
+        len(parts) != 2
+        or not all(parts)
+        or parsed.scheme != "trino"
+        or parsed.scheme != "presto"
```
evansd, line 103: I think you need
```py
or (parsed.scheme != "trinio" and parsed.scheme != "presto")
```
Or maybe neater to do it as
```py
or parsed.scheme not in ("trinio", "presto")
```

StevenMaude: I was going to use `not in`, then thought "why don't I make it easier to cleave off the change?" :man_facepalming:

StevenMaude:
I've had a go at allowing old and new settings to work side-by-side.

That said, it's the end of the week and warm, so it's possible I've missed something.

Style wise on the tests:
1. Is having warnings appear in the tests OK? Think it should be, but let me know if not.
2. Is there a better pytest-idiomatic way than what I've done to repeat the same tests with different URL schemes? The alternative to what I've done is duplicate the tests, but that seems a little unnecessary.

StevenMaude:
And maybe the `PRESTO` environment variables should take precedence for now, until `TRINO` is supported?

cohortextractor/trino_utils.py
```diff
@@ -54,16 +55,35 @@ def adapt_connection(conn, conn_params, env=os.environ):
     """
 
     session = requests.Session()
-    crt = env.get("PRESTO_TLS_CERT")
-    key = env.get("PRESTO_TLS_KEY")
```
evansd, line 58: This feels slightly overcomplicated to me as we're now duplicating various bits of code including writing temp files twice with different prefixes. I think you could have achieved the same thing with just:
```py
crt = env.get("TRINIO_TLS_CERT", default=env.get("PRESTO_TLS_CERT"))
key = env.get("TRINIO_TLS_KEY", default=env.get("PRESTO_TLS_KEY"))
```
Which would have been easier to back out.

But fine to leave as is, as ought to be short lived in any case.

evansd (approved):
All looks good. I made a few FYI comments, but nothing that needs acting on.

cohortextractor/trino_utils.py
StevenMaude: Here the intent is to guard against unexpected mixing and matching of different environment variables. For example, if we fallback to the `default` here, someone could supply **`TRINO`**`_TLS_CERT`, but specify **`PRESTO`**`_TLS_KEY`. This would work now. It would fail later when we only look for `TRINO_TLS_CERT` and `TRINO_TLS_KEY`.  This could catch "future me" (or "future someone else") out.

But, we're likely the only users actually specifying these variables. It may not matter too much to protect against this.

Agree that if it was to stay in the code perpetually, the duplication is more complex. But we should be able to revert back to the original structure that was there after removing the `PRESTO` variables.

On balance then, I'm happy to leave as is.

***

Relating to file prefixes, I'm not completely certain if the filenames are checked/loaded elsewhere. I don't think so from a quick search in `opensafely-core` for anything relevant.

tests/test_trino_utils.py
```diff
@@ -9,45 +9,57 @@
...
-        "user": "user",
-        "host": "host",
-        "port": 80,
-        "catalog": "catalog",
-        "schema": "schema",
-    }
+    urls = [
+        "trino://user:password@host:80/catalog/schema",
+        "presto://user:password@host:80/catalog/schema",
+    ]
```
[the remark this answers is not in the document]
StevenMaude: Thanks for confirming. I'll do this, because I want to know for my own curiosity, and not having used pytest before.

(Incidentally, the "loop through test cases" approach is [idiomatic Go](https://golang.org/doc/code#Testing), which is my defence here :wink:)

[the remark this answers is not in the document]
StevenMaude: What I was missing yesterday was that you must pass a variable set with `parameterize` into the test function itself.

StevenMaude:
https://github.com/opensafely-core/cohort-extractor/runs/3143275452?check_suite_focus=true
```
Head https://registry-1.docker.io/v2/starburstdata/starburst-enterprise/manifests/356-e.5: received unexpected HTTP status: 502 Bad Gateway
```

:man_facepalming:

PR #621, opened 2021-07-22
Feat: Implement outpatient_appointment_date for TPP

### LINKED ISSUE #492: Add out-patient appointment data to study definition
Issue for querying HES / SUS out-patient appointments data from the `OPA` tables in the OpenSAFELY-TPP database.

A [full data dictionary for the outpatient dataset in HES](https://nhs-prod.global.ssl.fastly.net/binaries/content/assets/website-assets/data-and-information/data-tools-and-services/data-services/hospital-episode-statistics/hes-data-dictionary/dd-op_v12.pdf) is available, though by the time this data gets to the OpenSAFELY-TPP database there's been some filtering and processing -- see [the database schema](https://github.com/opensafely/database-notebooks/blob/master/notebooks/database-schema.ipynb) to see what's available. 

Fields that are likely to be important:

* `Attendance_status` in the `OPA` table -- did the patient attend the appointment? 
* `Main_Specialty_Code` in the `OPA` table. This is the "the specialty under which the consultant is contracted".  One of around 100 possible codes eg ENT, obstetrics.
* `Treatment_Function_Code` in the `OPA` table. This is "the specialty in which the consultant was working during the period of care". Same set of codes as `Main_Specialty_Code`. I'm not sure the exact difference between this and `Main_Specialty_Code` though very often they're the same.
* `Primary_Procedure_code` in the `OPA_Proc` table. I.e., what procedure was undertaken, if any? These are OPCS-4 codes https://classbrowser.nhs.uk/#/book/OPCS-4.9. 
	* If there was a procedure, then in theory `Operation_Status` should be `1` from the `OPA` table. If none, eg consultation only, then `Operation_Status` should be `8`. But sometimes `Primary_Procedure_code` is `NULL` and `Operation_Status` is `8`. So not sure what's going on here.
	* There's also `Procedure_Code_2`. There are up to 23 secondary procedure codes in HES but this column only has at most one secondary procedure code.

The study definition variable will look something like this:

```
appointment_date = patients.outpatient_appointment_date(
	returning = "date",
	attended = True,
	with_these_treatment_codes = treatment_codelist,
	with_these_specialty_codes = specialty_codelist,
	with_these_procedure_codes = OPCS_codelist,
	on_or_after = "index_date",
	date_format="YYYY-MM-DD"
)
```

Potential other returning values that may be needed are `returning = "primary_procedure"` or `returning = "treatment_code"` or `returning = "specialty_code"` or `returning = "diagnosis"`, but these aren't immediately needed for the vaccine study.

## Use-case for vaccine study

The immediate use-case for this variable is to identify outpatient appointments that are likely to increase the chance of getting a covid vaccine. Principally because vaccines may have occurred or have been booked on the back of OP appointments. Essentially, we need to return the appointment date for a subset of appointments based on the type of appointment ("type" to be defined!). 

It's important to be able to exclude vaccination appointments, otherwise by using these appointments to predict vaccination we have a tautology. So any appointments involving vaccine-related OPCS codes must be explicitly excluded (eg `X44.3 Subcutaneous injection of vaccine`). Since only the primary procedure and one secondary procedure codes are available, then we may not catch all vaccine appointments.

We probably only need appointments that were actually attended (`Attendance_status==True`), although in theory it's possible that the vaccination was arranged at the time the appointment was _booked_ rather than when the appointment _occurred_.

## Multiple appointments on the same day
There's an added complexity to outpatient appointments given that multiple appointments on the same day are more likely, and info from all of them may be relevant and need to be queried / returned. This is tricky because it's not currently possible to return event-level data (ie, one row per patient per event), and `on_or_after`-type queries for getting consecutive event dates returned across multiple columns can't (?) be used to return two distinct events occurring on the same day.

Since appointment time isn't known (`Appointment_Date` is given as `YYY-MM-DD`) I think we'd have use the `OPA_Ident` appointment ID column instead. So for example `on_or_after="appointment_date_1"` internally creates two conditions: `appointment_date_2 >= appointment_date_1 AND appointment_ID_2 != appointment_ID_1`. But it's probably a mega-faff since `"appointment_date_1"` is just a date/string by the time it's passed to `on_or_after` (it doesn't additionally contain patient or appointment ID info). 

This sort of feature would be handy for hospital admissions and A&E attendances too.

madwort:
* fixes #492


cohortextractor/tpp_backend.py
```diff
@@ -2353,6 +2353,41 @@ def patients_with_healthcare_worker_flag_on_covid_vaccine_record(self, returning
...
+        returning="binary_flag",
+    ):
+        date_condition, date_joins = self.get_date_condition(
+            "OPA", "Appointment_Date", between
+        )
+
+        if returning == "binary_flag":
+            column_definition = "1"
+        elif returning == "date":
+            column_definition = "MIN(Appointment_Date)"
```
madwort, line 2373: other similar functions have parameters like "return first or return last result" to make this more flexible

StevenMaude (approved):
# Overall

As far as I can follow the changes, this looks good to me. **However, I have very little experience of this codebase and data.** Also agree that we should try and get the changes as they are merged in, then have extensions to this added later in a bigger PR.

As @madwort and I discussed, we think someone with more cohort-extractor experience should also review new functions for researchers that we're adding, e.g. @evansd or @inglesp.

Feedback from a researcher would also be very useful, although maybe this is post-merge, where someone can more easily try this out.

# Some comments/thoughts (feel free to ignore)

* I'd replace the few uses of "out-patient" with "outpatient" for consistency.
* You could consider referencing the original issue in the docstring for reference. I think the context is useful for interested readers trying to understand more of why the code is why the way it is; in case they have queries about code assignments or design decisions.
  * The code doesn't seem to routinely do this, but there are occasional examples (e.g. `admitted_to_hospital()` in `patients.py`). I think it's a good idea: the link is visible to anyone reading the docstring-based documentation on the OpenSAFELY documentation site.
* As @sebbacon has mentioned elsewhere, we should add this to the news page on `docs.opensafely.org`, once we're happy with them.

# A question

* How does having multiple different outpatient episodes for different treatments affect things here? It seems plausible it could happen. Does this need adding to the tests?

madwort:
> How does having multiple different outpatient episodes for different treatments affect things here? It seems plausible it could happen. Does this need adding to the tests?

Very good question - the issue ticket suggested outputting multiple events per patient, but the cohort-extractor framework appears to actively fight against that idea (by rejecting outputs with multiple rows per `Patient_ID`) - so my expectation is that end-users will typically filter based on treatment codes & time, meaning reducing the number of patients with multiple episodes in the output. In terms of the tests - the dummy data set has patients with multiple appointments, which I think will test this - but I could add some more treatment codes to verify this.

PR #622, opened 2021-07-23
fix: Fix booleans in binary format dummy data

evansd:
We previously round-tripped everything via CSV, this worked OK for
everything except booleans because: `bool("0") == True`


inglesp (approved):
Nicely done, and thanks for improving the comment.

PR #623, opened 2021-07-27
feat: Revert to Starburst with Presto

StevenMaude:
This PR reverts #619 which changed our integration with Starburst to use Trino. See #619 for more details. 

This PR is required because EMIS reverted back to using Starburst with Presto, albeit presumably the [final release](https://hub.docker.com/r/starburstdata/presto/tags?page=1&ordering=last_updated) with Presto available (2c6852a1b625a7796d28c771d43d4819cf70c35a).

This downgrade broke our check of the EMIS Starburst version, and is causing our CI to fail again.


StevenMaude:
Other notes here:
* I've restored the original branch from #619 so that it can be pull requested again in future. The changes there should be good, just that we need to wait for EMIS to upgrade again.
* Not sure what the conventional commit/semantic versioning interpretation is here. It's odd to have a revert be a feature, but if it is merely labelled `fix/chore`, it will only be a point release, which feels a bit off to me. Suggestions welcome.

madwort:
> Not sure what the conventional commit/semantic versioning interpretation is here. It's odd to have a revert be a feature. But if it is merely labelled fix/chore, it will only be a point release, which feels a bit off to me. Suggestions welcome.

Hmmm, it does sound a bit weird when you put it like that, but for me if you put it like "we supported dependency v1, this PR enables support for dependency v2" it does sound like a feature. However, should #619 have been `feat!:` instead of `feat:`, because it's a breaking change?

madwort (approved):

StevenMaude:
@madwort:

> However, should #619 have been `feat!:` instead of `feat:`, because it's a breaking change?

Possibly, yes…!

I did think that it could have been a major version bump after the release was made. (Although this conventional commit stuff is new to me: I didn't know the `!` convention for making that happen.)

However, it's not the "researcher API" that's changing, just the internal details of the database that you can connect to. A researcher wouldn't need to update their research code at all for this change. So I didn't have a good answer. It depends who the version number is aimed at: developers deploying changes to backends, researchers writing code for studies, or both.

If anyone has a strong opinion on this for when we eventually switch back to Trino again, I'm fine to go with that.

PR #624, opened 2021-07-27
ci: Add regular version check for Starburst

### LINKED ISSUE #620: Move the EMIS version check to a scheduled workflow
As opposed to running in the tests.

Suggested by [@evansd](https://github.com/opensafely-core/cohort-extractor/pull/593#issuecomment-865036536) and raised again by him today. We should check on a regular schedule: otherwise, this check only occurs when anyone is writing new code. EMIS might change the version in the interim.

StevenMaude:
Fixes #620.

Check the Starburst version running on EMIS, daily every weekday.

### DIFF
--- a/.github/workflows/emis_trino_version_check.yaml
+++ b/.github/workflows/emis_trino_version_check.yaml
@@ -1,0 +1,29 @@
+name: Check Trino version on EMIS and in docker-compose.yml match
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 8 * * 1-5'
+
+jobs:
+  check-starburst-version:
+    name: Check Starburst version
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+      - name: Install
+        run: |
+          set -xe
+          python -m pip install --upgrade pip
+          python -m venv venv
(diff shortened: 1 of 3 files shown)


StevenMaude:
This does fix #620, but might want slight reworking — the test filename gets renamed, and "Trino" should be "Presto" — very slightly based if #623 is merged. Likewise, the CI here fails because we're currently on the "wrong" version.

See an [example run](https://github.com/StevenMaude/cohort-extractor/runs/3174073993?check_suite_focus=true) of it working.

Also might not want running within the tests themselves, which might be accomplished by:
* separating the version check from the rest of the tests, and, for normal testing, run `pytest` in the directory that contains all but the version check test; 
* using `mark` and checking e.g. an environment variable

.github/workflows/emis_trino_version_check.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+      - name: Install
+        run: |
+          set -xe
+          python -m pip install --upgrade pip
+          python -m venv venv
+          # Save time by only installing the necessary requirements for the single test.
```
madwort, line 23: nifty!

evansd (approved):
LGTM

StevenMaude:
This will need slight changes — to change back to Presto for now — following #623.

madwort (approved):

PR #625, opened 2021-07-28
feat: Upgrade to Starburst Trino again

StevenMaude:
We originally upgraded to Starburst Trino in #619, and then backed out that change in #623.

Once EMIS have upgraded to Starburst Trino, we should merge this PR to reintroduce the changes in #619.

There is an additional revert of a commit in #624, necessary as that introduced a workflow and small test change that both referred to "Presto".

TODO:
- [x] Wait for EMIS to upgrade to Starburst Trino again.
- [x] Rebase this branch on main and fix any merge conflicts.
- [x] Check and update the version of Starburst in `docker-compose.yml` to match what EMIS are then using.
- [x] Probably worth checking if the version of the trino-python-client here needs bumping. From the [commit history and pattern of sporadic releases](https://github.com/trinodb/trino-python-client), the Trino Python client looks fairly stable and does not have a huge amount of development.
- [x] Confirm EMIS are sticking with the upgrade this time.


StevenMaude:
Today I updated this PR as I spotted a [failing EMIS Presto/Trino version check](https://github.com/opensafely-core/cohort-extractor/actions/runs/1265019934).

Before merging, we should confirm that EMIS are sticking with this upgrade.

evansd (approved):
:+1: This all looks good, based on what I've seen before.

Once deployed we should look at updating the config in @bloodearnest's configuration script to use the new keys (although no urgency because of the backwards compatibility code) and then make sure we update the config in the live EMIS environment.

PR #629, opened 2021-08-04
feat: Implement referral_request_received_date filter

madwort:
* requires #621 to be merged first
* filter outpatient_appointment_date for TPP by `Referral_Request_Received_Date`
* TODO: verify with @HelenCEBM that this is what she intended by this feature

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -2150,6 +2150,7 @@ def outpatient_appointment_date(
     is_first_attendance=None,
     with_these_treatment_function_codes=None,
     with_these_procedures=None,
+    referral_request_received_date=None,
     on_or_after=None,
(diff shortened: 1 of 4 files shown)


HelenCEBM:
HI @madwort 
I think the referral request date would need to accept a range of dates or on_or_before/on_or_after. The usefulness of this field could be to ascertain how long someone was waiting for their first appointment, or how long their period of care has been ongoing (eg someone with an original referral date 10 years ago would appear to be suffering with the same condition for that time). It will need to accept relative dates. If this is difficult we can drop it as the use cases will be minimal (but it might come up as a necessity at some point!).

madwort:
closing because this is insufficient & probably not hugely useful as a base for others to work from

PR #631, opened 2021-08-04
fix: Configure Presto timeout as per MS SQL

StevenMaude:
The `wait_for_presto_to_be_ready()` function is analogous to
`wait_for_mssql_to_be_ready()`. However, it is missing this step, as
included in this commit, of configuring the timeout.

The current use of both of these functions is in the tests, where the
`timeout` parameter for each function is set by the environment variable
`CONNECTION_RETRY_TIMEOUT`. This environment variable is retrieved by
`os.environ.get()` which means it will be `None` if not set, or a string
if present.

(There is the slight peculiarity of converting a user-supplied timeout
with `float`, while the default literal will be an `int`, but this line
is in the other function, so this is left unchanged.)


StevenMaude:
This change fixes the `test_emis_backend` tests for me, running via `./run.sh pytest` with [rootless Docker](https://docs.docker.com/engine/security/rootless/), on my slow-ish laptop. Otherwise, I get errors relating to Presto not being ready, e.g.

```
prestodb.exceptions.PrestoQueryError: PrestoQueryError(type=INTERNAL_ERROR, name=SERVER_STARTING_UP, message="Presto server is still initializing", query_id=20210803_171048_00000_g9rb9)
```

evansd (approved):
Thanks @StevenMaude

Happy with this as is, as you say it just replicates what we've already got in the mssql function. But I think actually the mssql function is wrong and what we want instead is:
```py
timeout = float(os.environ.get("CONNECTION_RETRY_TIMEOUT", "60"))
```
on these two lines:
https://github.com/opensafely-core/cohort-extractor/blob/09df42d22b37d17ff25392df43a51d0ca0a548af/tests/tpp_backend_setup.py#L27
https://github.com/opensafely-core/cohort-extractor/blob/09df42d22b37d17ff25392df43a51d0ca0a548af/tests/emis_backend_setup.py#L52

Otherwise it feels to me like the casting and default setting are happening at the wrong layer.

StevenMaude:
@evansd: I think your suggestion is more sensible; I'll revise :+1:

evansd (approved):

PR #633, opened 2021-08-04
feat: clean-up covariate return type definitions

madwort:
* order alphabetically
* ordering reveals a typo, so remove


cohortextractor/process_covariate_definitions.py
```diff
@@ -334,46 +334,45 @@ def _type_from_return_value(self, returning):
...
             "date": "date",
             "date_admitted": "date",
+            "date_arrived": "date",
             "date_discharged": "date",
             "date_of_death": "date",
+            "days_in_critical_care": "str",
+            "discharge_destination": "str",
+            "duration_of_elective_wait": "str",
+            "group_16": "str",
+            "group_6": "str",
```
madwort, line 353: this ordering isn't enforced, we could order numerically here rather than lexicographically

PR #634, opened 2021-08-05
Treat NaN values as being null

### LINKED ISSUE #627: Error generating categorical variables
Generating categorical variables (I presume) is throwing [this error](https://github.com/opensafely/long-covid/runs/3219046467).  I have also seen this [here](https://github.com/opensafely/SRO-Measures/runs/3172682127). This is potentially related to #622.

evansd:
In Pandas, all null values should be given a category index of -1.
Previously we used a Python truthiness test which treated NaNs as being
non-null meaning we tried to create a new category with the value `NaN`
which blows up.

Closes #627

### DIFF
--- a/tests/test_study_definition.py
+++ b/tests/test_study_definition.py
@@ -44,7 +44,7 @@ def test_create_dummy_data_works_without_database_url(tmp_path, monkeypatch):
 
 @pytest.mark.parametrize("file_format", SUPPORTED_FILE_FORMATS)
 def test_to_file_with_expectations_population(tmp_path, file_format):
-    cl = codelist(["12345"], system="snomed")
+    cl = codelist([("12345", "foo"), ("67890", "bar")], system="snomed")
     study = StudyDefinition(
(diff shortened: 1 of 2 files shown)


tests/test_study_definition.py
```diff
@@ -84,6 +84,15 @@ def test_to_file_with_expectations_population(tmp_path, file_format):
...
         ),
+        incomplete_categories=patients.with_these_clinical_events(
+            cl,
+            returning="category",
+            return_expectations={
+                "category": {"ratios": {"foo": 0.5, "bar": 0.5}},
+                # Half the values here should be null
+                "incidence": 0.5,
+            },
+        ),
```
inglesp, line 95: What does this actually test?

inglesp (approved):

tests/test_study_definition.py
evansd: Without the "fix:" commit this test errors out, with it the test passes.

PR #638, opened 2021-08-11
fix: #636 minimum_of fails expectations generation with **extra_columns

### LINKED ISSUE #636: patients.minimum_of() fails dummy data generation with **extra_columns
e.g.

```
death_date=patients.minimum_of(
        primary_care_death_date=patients.with_death_recorded_in_primary_care(
            on_or_after="index_date",
            returning="date_of_death",
            date_format="YYYY-MM-DD",
            return_expectations={
                "date": {"earliest" : "index_date"},
                "rate" : "exponential_increase"
            }
        ),
        cpns_death_date=patients.with_death_recorded_in_cpns(
            on_or_after="index_date",
            returning="date_of_death",
            date_format="YYYY-MM-DD",
            return_expectations={
                "date": {"earliest" : "index_date"},
                "rate" : "exponential_increase"
            }
        ),
        ons_died_from_any_cause_date=patients.died_from_any_cause(
            on_or_after="index_date",
            returning="date_of_death",
            date_format="YYYY-MM-DD",
            return_expectations={
                "date": {"earliest" : "index_date"},
                "rate" : "exponential_increase"
            }
        )
    ),
```

produces 

```
[...]
File "/app/cohortextractor/study_definition.py", line 359, in make_df_from_expectations
         columns = df[list(definition_args["column_names"])]
       File "/usr/local/lib/python3.8/dist-packages/pandas/core/frame.py", line 2806, in __getitem__
         indexer = self.loc._get_listlike_indexer(key, axis=1, raise_missing=True)[1]
       File "/usr/local/lib/python3.8/dist-packages/pandas/core/indexing.py", line 1551, in _get_listlike_indexer
         self._validate_read_indexer(
       File "/usr/local/lib/python3.8/dist-packages/pandas/core/indexing.py", line 1639, in _validate_read_indexer
         raise KeyError(f"None of [{key}] are in the [{axis_name}]")
     KeyError: "None of [Index(['primary_care_death_date', 'cpns_death_date',\n       'ons_died_from_any_cause_date'],\n      dtype='object')] are in the [columns]"
```

[Relevant slack thread](https://ebmdatalab.slack.com/archives/C33TWNQ1J/p1628255196051700)

Jongmassey:
- add `**extra_columns` column definition to min and max variables in `test_make_df_from_expectations_with_aggregate_of()`
- create columns in for aggregation but not output within `make_df_from_expectations()`:
  - make temporary inner df for extra columns
  - cast columns of temporary df to dtype of first column of `*columns` arg of aggregate column
    - *hacky workaround to fact that date columns in temp df will be converted to string as last step of generation*
  - concat temporary df with `*columns`-based columns
  - pass these concatenated columns to existing aggregation steps

### DIFF
--- a/cohortextractor/study_definition.py
+++ b/cohortextractor/study_definition.py
@@ -351,32 +351,9 @@ class StudyDefinition:
                 method_name = "max"
             else:
                 raise ValueError(f"Unsupported aggregate function '{fn}'")
-            extra_columns = {
-                k: v
-                for (k, v) in self.covariate_definitions.items()
-                if k in list(definition_args["column_names"]) and v[1]["hidden"]
-            }
-            if extra_columns:
-                extra_study = StudyDefinition(
-                    self._original_covariates["population"],
-                    self.default_expectations,
-                    self.index_date,
-                    **extra_columns,
-                )
-                extra_df = extra_study.make_df_from_expectations(population)
-                non_extra_columns = [
-                    c
-                    for c in list(definition_args["column_names"])
-                    if c not in extra_columns.keys()
-                ]
-                dt = df.dtypes[non_extra_columns[0]]
-                extra_df = extra_df.astype(dt)
-                extra_df = pd.concat(
-                    [df[non_extra_columns], extra_df], ignore_index=True
-                )
-                columns = extra_df[list(definition_args["column_names"])]
-            else:
-                columns = df[list(definition_args["column_names"])]
+            columns = self.get_columns_for_aggregation(
+                population, df, list(definition_args["column_names"])
+            )
             aggregate_method = getattr(columns, method_name)
             df[colname] = aggregate_method(axis=1)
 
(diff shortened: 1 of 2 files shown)


cohortextractor/study_definition.py
```diff
@@ -428,7 +453,7 @@ def check_date_expectations_defined(self, colname, kwargs):
 
 
 def merge(dict1, dict2):
-    """ Return a new dictionary by merging two dictionaries recursively. """
```
Jongmassey, line 431: this was `black` not me!

evansd: Ah, it looks like this is new behaviour from black and you must be running a different version from the one in `requirements.txt`. That's fine though, it's a sensible enough change and we should be using a more recent version of black anyway.

evansd (changes requested):
Thanks for doing this! In one sense this is definitely not The Right Way to solve this problem, but given that:
 * I can understand how and why it works;
 * the whole dummy data system needs a massive overhaul; and
 * the whole cohortextractor is being replaced anyway

I think it's fine to live with the hack!

I like though to pull it out into a separate method, something like:
```py
columns = self.get_columns_for_aggregation(df, definition_args["column_name"], population)
```
which will at least consolidate the hack into one place.

A couple of small issues:

1. The `fix: ` prefix actually needs to go on a commit title, not just the PR title. The "Semantic Pull Request" check gives a [misleading](https://github.com/opensafely-core/cohort-extractor/issues/242) green tick here but we never get round to fixing that.
2.  While you doing that you might as well squash this commit and the black reformat together.
3. There's a convention to keep commit titles (and by extension PR titles which become merge commits) to under 72 chars. Nothing terrible will happen if you exceed this but it can make scanning commit logs harder.

It really is great that you've fixed this though. Thanks again!

evansd (approved):
:+1:

PR #641, opened 2021-08-17
fix: msoa EMIS backend inconsistency

alexwalkercebm:
The [documentation for MSOA](https://docs.opensafely.org/study-def-variables/#cohortextractor.patients.registered_practice_as_of) says that `"msoa"` should be used to specify it. This works in TPP, (which also supports using `"msoa_code"`), but in EMIS only `"msoa_code"` works. This has tripped people up a couple of times.

This PR should make the EMIS backend consistent with TPP: https://github.com/opensafely-core/cohort-extractor/blob/4a339e7fb67a980ee5e9719008afbf25b43c0052/cohortextractor/tpp_backend.py#L1292-L1295

### DIFF
--- a/cohortextractor/emis_backend.py
+++ b/cohortextractor/emis_backend.py
@@ -1121,7 +1121,9 @@ class EMISBackend:
 
         if returning == "stp_code":
             column = "stp_code"
-        elif returning == "msoa_code":
+        # "msoa" is the correct option here, "msoa_code" is supported for
+        # backwards compatibility
+        elif returning in ("msoa", "msoa_code"):
             column = "msoa"
         elif returning == "nuts1_region_name":
             column = "english_region_name"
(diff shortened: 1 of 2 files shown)


sebbacon (approved):
I think this is fine, but FYI if I really cared about the EMIS backend (I consider it experimental, and backwards compatibility isn't really that important to test) I would add tests for the old behaviour, too.

If you check [the merge that introduced the new behaviour](https://github.com/opensafely-core/cohort-extractor/commit/d8cfcfe4c90c95e6d21669990960be9fa5decada) in the TPP lines you quoted, you'll see how that was done.

Also FYI I found that commit by doing a `blame` (link at top right of source code view) on the TPP code. That shows the last commit that touched those lines.

alexwalkercebm:
Thanks, I missed that the old behaviour was tested in TPP. Have added to EMIS for completeness.

sebbacon:
I've merged but you'll need to request someone deploys in EMIS for you in #tech

PR #642, opened 2021-08-18
docs: Remove redundant value type from docstrings

### LINKED ISSUE #364: remove default value text in Args docstring section
For instance "the default value is `none`" is not needed since it's pulled from the function def and listed in the argument table

StevenMaude:
Fixes #364.

This information is already in the function definition. In the
documentation, we show the default value as a separate column. This
means that we get duplication where we say "The default value is `None`"
and then show the default is `None`.

This is also brittle in that someone might modify a default, but forget
to update the docstring, or vice versa.

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -269,13 +269,13 @@ def most_recent_bmi(
 
     Args:
         on_or_before: date of interest as a string with the format `YYYY-MM-DD`. Filters results to measurements
-            on or before the given date. The default value is `None`.
+            on or before the given date.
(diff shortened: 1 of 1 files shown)


StevenMaude:
This was requested in #364. I'm open minded about this, so happy to be persuaded it's not required.

It does remove the duplication you get in the docs: [example](https://docs.opensafely.org/study-def-variables/#cohortextractor.patients.date_deregistered_from_all_supported_practices)

default-value-duplication

sebbacon (approved):

PR #644, opened 2021-08-26
feat: implement consultation medium return values for outpatient appointments

madwort:
* more functionality from https://github.com/opensafely-core/cohort-extractor/issues/492
* return `consultation_medium_used`
* return either first or last match
* also improve `date` return values to return first or last

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -2153,6 +2153,7 @@ def outpatient_appointment_date(
     on_or_after=None,
     between=None,
     date_format="YYYY-MM-DD",
+    find_first_match_in_period=None,
 ):
     """
     Return when the patient had an outpatient appointment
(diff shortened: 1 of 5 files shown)


cohortextractor/patients.py
```diff
@@ -2165,6 +2165,7 @@ def outpatient_appointment_date(
             * `binary_flag`: indicates if they have had an outpatient appointment or not
             * `date`: earliest date of outpatient appointment within the specified period
             * `number_of_matches_in_period`: number of outpatient appointments in period
+            * `consultation_medium_used`: consultation medium code for the earliest outpatient appointment within the specified period
```
inglesp, line 2168: Could we point to where the possible codes are documented?

inglesp (approved):

cohortextractor/patients.py
madwort: asked at https://github.com/opensafely-core/cohort-extractor/issues/492#issuecomment-910365078

madwort: added a link

Issue #563, opened 2021-06-03
Return "number_of_matches_in_period" for SGSS tests


wjchulme:
The number of PCR or LFT tests taken indirectly tells us about a person's health seeking behaviour / exposure / occupation / etc. This will be useful for the comparative vaccine effectiveness analysis to check balance between matched pfizer and az recipients.

CarolineMorton:
I think this is increasingly going to be of interest in studies in the near future. One thing that occurs to me is, is the data available to us to query?

wjchulme:
We have the `AllTests` SGSS tables `https://reports.opensafely.org/reports/opensafely-tpp-database-schema/` which in theory contains every test, positive and negative, reported via SGSS. But the consistency of data flows across different pillars, PCR/LFTs, etc is not a sure thing, so it would have to be used with care.

sebbacon:
@wjchulme, what does "used with care" mean in practice? Clear documentation? Or something we can assert in code?

If the former, it would probably be useful for the implementor to have the suggested documentation up-front; in any case, it'll need writing at some point, and I'm prioritising this to be picked up in the next 4 weeks, so if you can sketch it out, that'd be useful. Thanks!

wjchulme:
There are a lot of unknowns with the values that are returned by `patients.with_test_result_in_sgss`. In particular test provenance and test coverage. So by "used with care" I just mean that the value returned by counting the number of tests in a given period my be difficult to interpret or misleading without more information. Basically, it makes it less useful.

A note in the documentation will help to make these caveats clear.

What will help even more is being able to _filter_ tests by LFT/PCR and pillar / lab type, so that test provenance can be better understood and the coverage of testing data over time can be more easily gauged. Pillar/lab type is requested in #248. Currently, the only way to get LFT/PCR info is [by `returning` (not filtering) the `case_category` of the first-ever positive test](https://docs.opensafely.org/study-def-variables/#cohortextractor.patients.with_test_result_in_sgss--case-category-type-of-test-used), but there's also the `LFT_flag` variable in the `SGSS_AllTests_Positive` and `SGSS_AllTests_Negative` tables which should have this information for all tests.

acagreen17:
Having the number of PCR or LFT tests would also be really useful for the descriptive breakthrough COVID stuff - it could help to explain some of the differences in rates that we are seeing (i.e., higher rates in groups such as health care workers/immunocompromised might be due to higher rates of testing).

sebbacon:
[...]

PR #647, opened 2021-09-08
Support returning test counts from SGSS

evansd:
### Background
Patients can have multiple COVID tests and knowing how many they have had tells us something about their health seeking behaviour and propensity for testing. At present, we can only get information about 1 test at a time. 

See #563 for more Background information.

### What this PR does?
This PR changes the code in `tpp_backend.py` to get a count of all the times that a patient has appeared in one or both tables called `SGSS_AllTests_Negative` and `SGSS_AllTests_Positive`. It then sums these counts together and returns this value. 

It adds a new `returning` option which is `returning="number_of_matches_in_period"` (This is the same API as other parts of study definition that return number of matches such as `with_clinical_events()`). 

The time period can be restricted by using the `on_or_before` or `between` dates as per the existing API. 

See Tests for examples. 

I have added an Error to be raised if `returning="number_of_matches_in_period` with `restrict_to_earliest_specimen_date=True`. 

### What this PR does not do?
It does not yet filter by type of test, such as PCR or Lateral Flow Test. I would like to be done in a separate PR to make it clearer.

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -1589,6 +1589,15 @@ def with_test_result_in_sgss(
 
     Possible values are "", "Y", "N".
 
+    ### Number of Tests
+
+    The `returning="number_of_matches_in_period"` option (only available in the "All Tests" data)
+    returns a count of the number of tests a patient has had in the defined time period.
+
+    It is used with `test_result` which must be set as "positive", "negative" or "any".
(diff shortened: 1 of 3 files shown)


evansd:
@CarolineMorton Thanks for the docs stuff. I can't approve this PR though as I'm still considered as the author, can you do the honours?

CarolineMorton (approved):

PR #648, opened 2021-09-08
GermDefence RCT

### LINKED ISSUE #599: RCT dataset support (GermDefence)
New collaborators from Bristol are joining their _GermDefence_ RCT dataset (which was exposing GPs / practices to a website-based intervention) into primary care data, to investigate health outcomes (the incidence of respiratory tract infections, COVID-19 diagnoses, COVID-19 symptoms, and gastrointestinal infections, and the number of primary care consultations and antibiotic prescriptions)

The raw data columns (27 of them) are pasted below.

They are only interested in data at a practice, not patient level. 

So with the study definition model they'd want something like `patients.with_germdefence_rct()`, then define one variable for each outcome. How would they get the intervention data? Would we add something to `registered_practice_as_of`, along the lines of registered_practice_as_of(date, returning='germdefence__n_pages_viewed_mean')`?

The novel problem to consider here is that this is one-off; no-one else will have a use-case for germdefence data (in fact, no-one else should be able to access it at all, ideally).  So baking it into the core study definition API seems very wrong.

We should also consider that longer term we might get more of these per-patient or per-practice variables that are only to be used for a single project. 

Maybe there is a generic study definition call we could define, for random stuff like this, e.g. `patients.with_custom_property(dataset='germdefence')` or whatever.

Or perhaps we'd implement this as a backend (e.g. `tpp-germdefence`).  And backends get their own subsets of study definition variables; and we implement an action to allow you to join data from different backends (assuming they're on the same server). The nice thing about this is we'd get to lock down access per-user.

columns:

```csv
code,
ccg_code,
listsize_Jan21,
deprivation_score,
intervention,
IntCon,
List_size,
IMD_rank,
IMD_decile,
MeanAge,
MedianAge,
Av_rooms_per_house,
Minority_ethnic_total,
n_times_visited_mean,
n_pages_viewed_mean,
total_visit_time_mean,
prop_engaged_visits,
n_engaged_visits_mean,
n_engaged_pages_viewed_mean_mean,
N_visits_practice,
group_mean_behaviour_mean,
group_mean_intention_mean,
N_completers_RI_behav,
N_completers_RI_intent,
hand_behav_practice_mean,
hand_intent_practice_mean,
N_completers_HW_behav,
N_goalsetting_completers_per_practice
```

iaindillingham:
Adds support for randomized controlled trials (RCTs), principally the GermDefence RCT. This PR is a WIP. Watch out for force-pushes ⚠️.

Closes #599


iaindillingham:
Thanks for taking this over @CarolineMorton. Next steps, in rough order:

* **Checkout the branch and run the tests.** Make sure everything's ✅ before starting 🙂.
* **Add the remaining properties.** For each, you'll need to add to *cohortextractor/process_covariate_definitions.py* and *cohortextractor/tpp_backend.py* and update the test in *tests/test_tpp_backend.py*. The property names are given in #599, but I'd log in to TPP and double-check that they're correct. It's also nice to have real values in the test (I've left a couple of `FIXME` comments) and you'll only get these from logging into TPP. I'd strongly encourage you to add one property at a time, checking that the tests pass.
* **Think about the syntax for the properties**; don't forget to ask @HelenCEBM as she's co-piloting. At present, each property is `rct__[trial name]__[property name]` e.g. `rct__germdefence__code`. The property should be valid SQL -- hence the dunder (`__`) -- but the constant `rct`, the trial name, and the property names are up for grabs.
* **Document the properties** in *cohortextractor/patients.py*.
* **Consider refactoring** the SQL queries in `patients_registered_practice_as_of`. I think the two new queries could possibly be refactored into a single query, giving one new query and one existing query; I'm unsure whether they could be refactored into the existing query. [Timebox](https://en.wikipedia.org/wiki/Timeboxing) this work; I can pick it up when I'm back.
* **Consider refactoring** the mapping in `patients_registered_practice_as_of`. I've left a comment. Again, timebox; I can pick it up when I'm back.

CarolineMorton:
Picking this up from @iaindillingham. Will go through current state of code and what @sebbacon  and I discussed today. Plan is for @evansd to review today as germdefence team are keen to crack on ASAP.  

> * **Checkout the branch and run the tests.** Make sure everything's ✅ before starting 🙂.

All working (I had some issues related to needing update Docker, assign more memory, ad rebuild containers but we got there in the end). 

> * **Add the remaining properties.** For each, you'll need to add to _cohortextractor/process_covariate_definitions.py_ and _cohortextractor/tpp_backend.py_ and update the test in _tests/test_tpp_backend.py_. The property names are given in [RCT dataset support (GermDefence) #599](https://github.com/opensafely-core/cohort-extractor/issues/599), but I'd log in to TPP and double-check that they're correct. It's also nice to have real values in the test (I've left a couple of `FIXME` comments) and you'll only get these from logging into TPP. I'd strongly encourage you to add one property at a time, checking that the tests pass.

All of these properties have now been added one by one and have associated tests. I have not been able to log into L2 at present due to being without my laptop but I have asked Helen to check for me. I should be able to log in tomorrow in any event. For now there are placeholder values available, and I have made a best guess about type (i.e. str, int...etc) based on the property name 

> * **Think about the syntax for the properties**; don't forget to ask @HelenCEBM as she's co-piloting. At present, each property is `rct__[trial name]__[property name]` e.g. `rct__germdefence__code`. The property should be valid SQL -- hence the dunder (`__`) -- but the constant `rct`, the trial name, and the property names are up for grabs.

I have been through this with Helen, and she thinks that it looks ok, and is simple enough to use. We discussed whether the `rct__` was really necessary but on balance, we decided to keep this incase we want to use this format to add some other practice level variable in the future. 

> * **Document the properties** in _cohortextractor/patients.py_.

This has been done. Once approved, I will be able to update the documentation repo as well. 

> * **Consider refactoring** the SQL queries in `patients_registered_practice_as_of`. I think the two new queries could possibly be refactored into a single query, giving one new query and one existing query; I'm unsure whether they could be refactored into the existing query. [Timebox](https://en.wikipedia.org/wiki/Timeboxing) this work; I can pick it up when I'm back.
> * **Consider refactoring** the mapping in `patients_registered_practice_as_of`. I've left a comment. Again, timebox; I can pick it up when I'm back.

@sebbacon and I discussed this today. Overall Seb feels this could be pushed down the line and is not key to deploying a version of this live so that the germ defence team can get started. The refactor queries can be done post-deployment.

evansd (approved):
I think this is fine to go now.

As we've previously discovered there are various bit so refactoring that could be done here, but they don't need to block merging of an urgent feature.

I was originally going to suggest changing the variable names to match our conventions better e.g `mean_age` instead of `MeanAge`. (I think that's the point of the `apps_to_db_property_name` dict, otherwise it's just mapping a bunch of values to themselves.) But on further consideration, maybe it's better to have these match what the people supplying the data originally used, as they're also going to be the people using it.

I think there are only two things that really need doing:
 * for completeness we should probably have a full list of the available values in the docs
 * one of the commits will need a `feat:` prefix so we generate a new release of cohortextractor

I can do both of these tomorrow quickly enough if that's easier.

evansd:
FYI @CarolineMorton I've just fixed an issue with the tests which would make them hang forever at the end of the run:
https://github.com/opensafely-core/cohort-extractor/pull/648/commits/23cde9d8f506003c068df02ce0d166802ef4b1ae

This would probably explain all the problems you were having trying to run the test locally.

CarolineMorton:
I have now checked the columns in the database, specifically in `ClusterRandomisedTrialDetail`. Of the properties discused from our [original ticket](https://github.com/opensafely-core/cohort-extractor/issues/599#issue-933751868), some are not available. See below:
```
code, --> does not exist in ClusterRandomisedTrialDetail
ccg_code,--> does not exist in ClusterRandomisedTrialDetail
listsize_Jan21,--> does not exist in ClusterRandomisedTrialDetail
deprivation_score, --> actually named deprivation_pctile
intervention,--> does not exist in ClusterRandomisedTrialDetail
IntCon,
List_size,--> does not exist in ClusterRandomisedTrialDetail
IMD_rank,--> does not exist in ClusterRandomisedTrialDetail
IMD_decile,
MeanAge,
MedianAge,
Av_rooms_per_house,
Minority_ethnic_total,
n_times_visited_mean,
n_pages_viewed_mean,
total_visit_time_mean,
prop_engaged_visits,
n_engaged_visits_mean,
n_engaged_pages_viewed_mean_mean,
N_visits_practice,
group_mean_behaviour_mean,
group_mean_intention_mean,
N_completers_RI_behav,
N_completers_RI_intent,
hand_behav_practice_mean,
hand_intent_practice_mean,
N_completers_HW_behav,
N_goalsetting_completers_per_practice
```

HelenCEBM:
I think there is one called Deprivation_rank or similar rather than IMD_rank

evansd:
For reference here is the full list of currently available properties:
Screenshot from 2021-09-17 11-00-11

cohortextractor/patients.py
```diff
@@ -753,6 +755,60 @@ def registered_practice_as_of(
...
+
+
+        ### Cluster RCTs
+
+        Clustered Randomised Control Trials (Cluster RCTs) that are clustered at practice level
+        are available. There is special syntax that signifies RCT. This is `rct__trialname__propertyname`
+        for example, if there was a cluster RCT called germdefence which has a property called
+        deprivation_pctile, a variable can be created by:
+
+            germdefence_intervention=patients.registered_practice_as_of(
```
HelenCEBM, line 767: probably would call this variable germdefence_practice_deprivation_pctile or similar

cohortextractor/patients.py
```diff
@@ -753,6 +755,60 @@ def registered_practice_as_of(
                     },
                 },
             )
+
+
+        ### Cluster RCTs
+
+        Clustered Randomised Control Trials (Cluster RCTs) that are clustered at practice level
+        are available. There is special syntax that signifies RCT. This is `rct__trialname__propertyname`
```
HelenCEBM, line 763: Could add "A series of data files supplied by the trialists will be imported into OpenSafely; this will indicate which practices are enrolled, their assignment to an intervention group, and any other relevant practice properties or data gathered as part of the RCT outside of OpenSafely (e.g. number of GPs/nurses, number of practice visits made)."

cohortextractor/patients.py
```diff
@@ -753,6 +755,60 @@ def registered_practice_as_of(
...
+                            "1": 0.5,
+                            "2": 0.5
+                        },
+                    },
+                },
+            )
+
+        Note that a double underscore is required between `rct` and `trial_name` and between `trial_name`
+        and `property` (in this case `deprivation_pctile`).
+
```
HelenCEBM, line 783: Not sure it's too obvious to point out: "The data resulting from the study definition will be at patient level as usual and therefore practice variables will be repeated many times for each practice, and should be aggregated in a later analysis step"

cohortextractor/patients.py
HelenCEBM: Also: "These RCT variables are only available for use by the researchers officially nominated by the responsible research group."

PR #651, opened 2021-09-15
fix: Fix bug generating dummy data for aggregates

evansd:
As described here:
https://github.com/opensafely/documentation/discussions/406

Specifically, a KeyError was triggered when:
 * `minimum_of`/`maximum_of` was used somewhere in the study with
    internal (or "hidden") columns
 * the `population` definition was an expression referencing other
   columns.

This is due to the somewhat hacky workaround we use for generating dummy
data for these aggregation functions. (No aspersions cast on @Jongmassey
for this: before it worked in _no_ cases, now it works in some cases; so
it's definitely a strict improvement.)

I think the right fix here is to refactor things so that we generate
dummy data for all columns, hidden or visible, in one go and then we
just remove the hidden columns before writing the data out. But the
simpler fix for now is just to hardcode the `population` definition here
to something unproblematic. It plays no actual role in this case, so it
doesn't matter what we set it to.

### DIFF
--- a/cohortextractor/study_definition.py
+++ b/cohortextractor/study_definition.py
@@ -374,9 +374,14 @@ class StudyDefinition:
         }
         if extra_columns:
             extra_study = StudyDefinition(
-                self._original_covariates["population"],
-                self.default_expectations,
-                self.index_date,
+                # It doesn't matter what the population is here, it's not
+                # relevant in generating the extra columns we need. But we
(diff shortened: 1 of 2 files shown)


inglesp (approved):
Filing under "if you can't say anything nice, don't say anything at all"

Jongmassey:
No hard feelings at all!

Issue #601, opened 2021-07-02
Add additional information for lab/test data


acagreen17:
This request has been brought about through an external OpenSAFELY project but if we are increasingly using test values it is likely that the additional information will be useful going forward.

**Use-case from our external OpenSAFELY user:**
_We are currently using OpenSafely to assist with COVID-19 recovery planning in the area of colorectal cancer. Quantitative faecal immunochemical testing (FIT) has been more widely deployed in primary care during the pandemic period. We aim to understand how its use has impacted colorectal cancer diagnostic pathways and diagnoses. At present, laboratory data in the primary care recorded can be queried using the with_these_clinical_events, returning dates and numeric data. However, for some tests like FIT, the result is often reported using an inequality sign (e.g. <10 or >200). This is important - <10 clearly has a different meaning to 10. For some results, there may just be a normal/abnormal result, which will be missed altogether by the conversion to a number._
 
_I would therefore propose that the following are made available via the study definition interface:_
_1. Any inequality sign if present in the original data (‘>’, ‘<’ or none)_
_2. An abnormal/normal flag (with some indication as to whether this flag is simply not present; i.e. a TRUE/FALSE/missing value rather than a simple Boolean value)_
_3. As an alternative to (2), and depending on how the data are stored, a normal/low/critically low/high/critically high/missing value_
_4. The reference range_

To back up this use-case, there are some histograms of the FIT results, which do demonstrate the disproportionately large number of 0s/10s (numbers in the 000's). It’s worth noting that the common values vary by STP. In some STPs it’s 4, others 7, others 10. Some have lots of zeros (which is the default for missing values, but could potentially also represent actually a textual result?) Note, these can be re-formatted and redacted as needed.

evansd:
I think we have the raw data to support something like this in TPP, via the `CodedEventRange` table:

ColumnName | ColumnType | MaxLength | Precision | Scale | IsNullable
-- | -- | -- | -- | -- | --
Patient_ID | int | 4 | 10 | 0 | False
CodedEvent_ID | bigint | 8 | 19 | 0 | False
CodedEventRange_ID | bigint | 8 | 19 | 0 | False
Consultation_ID | bigint | 8 | 19 | 0 | False
LowerBound | real | 4 | 24 | 0 | False
UpperBound | real | 4 | 24 | 0 | False
Comparator | tinyint | 1 | 3 | 0 | False

[...]

PR #655, opened 2021-09-22
Support comparators and reference ranges

evansd:
This adds supports for fetching the comparator (`<`, `>=` etc) and the reference range associated with a lab result as described in #601.

This is currently TPP-only although from early sight of proposed EMIS schema additions it should be fairy straightforward to support the same API there.

We've discussed adding a function `is_within_reference_range()` which returns a "Yes", "No", "Can't say" for a given lab result based on the value, comparator and reference range bounds. This is certainly possible but there's some fiddly logic here which will require thought and careful testing and we don't (I think) want to block forward movement on making the raw data available until this feature is ready.

After merge, this docs PR will need to be merged:
https://github.com/opensafely/documentation/pull/422


iaindillingham (approved):
I more-or-less follow, and it seems reasonable, so ✅. As with previous changes to cohort extractor, it's the construction of the query that I find hardest to understand (although this is probably hindered by me being unfamiliar with the DB schema).

cohortextractor/process_covariate_definitions.py
```diff
@@ -16,7 +16,9 @@ def process_covariate_definitions(covariate_definitions):
     covariate_definitions = apply_compatibility_fixes_for_include_date(
         covariate_definitions
     )
-    covariate_definitions = add_include_date_flags_to_columns(covariate_definitions)
+    covariate_definitions = ensure_source_columns_genereate_required_values(
```
StevenMaude, line 19: Only looking at this, because I looked at https://github.com/opensafely/documentation/pull/422/ as I watch that repository, and it referenced this PR here.

Guessing there's no intentional reason behind the misspelling of `generate`?

sebbacon:
> We've discussed adding a function is_within_reference_range() which returns a "Yes", "No", "Can't say" for a given lab result based on the value, comparator and reference range bounds. This is certainly possible but there's some fiddly logic here which will require thought and careful testing and we don't (I think) want to block forward movement on making the raw data available until this feature is ready.

I'd like to discuss this further. One argument is that precisely *because* it's fiddly is the reason we *should* be considering implementing now.

@HelenCEBM understands pathology data well, I'd suggest a call with her, and me, to discuss. I could find 10 mins today.

I think this is a useful question to illuminate the more general question about how we hope to specify the next generation of study definitions.

cohortextractor/process_covariate_definitions.py
evansd: Thanks @StevenMaude, my vim autocompletion has a habit of propagating typos once they're made!

PR #656, opened 2021-09-23
fix: Allow using dates as categories

evansd:
This fixes the issue identified here:
https://github.com/opensafely/documentation/discussions/408#discussioncomment-1334495

Previously, it was not possible to use dates as categories because:
 * the system treated them as strings and wouldn't let you use them in
   places where it expected a date; and
 * our "reformat anything that looks like a date" hack would step in and
   reformat these dates in a way that prevented them being used
   elsewhere where ISO strings were expected.

Now they behave as expected so that you can use a column defined with
`categorised_as` as the input to another function which expects a date.

For example:
```py
study = StudyDefinition(
    population=patients.all(),
    eligible_date=patients.categorised_as(
        {
            "2020-04-14": "age >= 80",
            "2020-06-16": "age >= 70 AND age < 80",
            "2020-08-18": "DEFAULT",
        },
        age=patients.age_as_of("2020-01-01"),
    ),
    first_event_after_eligible=patients.with_these_clinical_events(
        codelist(["foo"], system="ctv3"),
        on_or_after="eligible_date",
        returning="date",
        date_format="YYYY-MM-DD",
    ),
)
```

### DIFF
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -344,6 +344,24 @@ class TPPBackend:
             raise ValueError(
                 "At least one category must be given the definition 'DEFAULT'"
             )
+
+        # We pass the `reformat_dates=False` option to the quoting function
+        # here to preserve date-like strings in the user-supplied ISO format,
+        # as opposed to converting to MSSQL-friendly format which we otherwise
+        # do by default. The distinction here is between dates used as inputs
+        # and dates used as outputs. We always need to supply input dates in a
+        # format which will be consistently parsed by MSSQL indepent of locale
+        # settings (amazingly, this does not include ISO format). But we always
+        # want the dates we output to be in ISO format. In almost all cases
+        # user-supplied dates are functioning as inputs, which is why
+        # reformatting is our default behaviour; but in this context they
+        # function as outputs so we need to do something different. (Note that
+        # if these outputs later go on to be supplied as inputs to other
+        # functions then the `date_expressions` code will ensure they are
+        # parsed correctly.)
+        def quote_category(value):
+            return quote(value, reformat_dates=False)
+
         # For each column already defined, determine its corresponding "empty"
         # value (i.e. the default value for that column's type). This allows us
         # to support implicit boolean conversion because we know what the
@@ -361,12 +379,15 @@ class TPPBackend:
             formatted_expression, names_used = format_expression(
                 expression, other_columns, empty_value_map=empty_value_map
             )
-            clauses.append(f"WHEN ({formatted_expression}) THEN {quote(category)}")
+            clauses.append(
+                f"WHEN ({formatted_expression}) THEN {quote_category(category)}"
+            )
             # Record all the source tables used in evaluating the expression
             for name in names_used:
                 tables_used.update(other_columns[name].source_tables)
+
         return ColumnExpression(
-            f"CASE {' '.join(clauses)} ELSE {quote(default_value)} END",
+            f"CASE {' '.join(clauses)} ELSE {quote_category(default_value)} END",
             type=column_type,
             # Note, confusingly, this is not the same as the `default_value`
             # used above. Above it refers to the value the case-expression will
@@ -375,6 +396,7 @@ class TPPBackend:
             # apart from bools and ints where it's zero.
             default_value=self.get_default_value_for_type(column_type),
             source_tables=list(tables_used),
+            date_format="YYYY-MM-DD" if column_type == "date" else None,
         )
 
     def get_aggregate_expression(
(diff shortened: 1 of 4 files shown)


sebbacon:
Neat. I wonder how we could document it best so it's discoverable by others. Would this pattern have a well-known name to epi researchers?

wjchulme:
@sebbacon I don't think there's a common name for this. It's really just allowing a bit more flexibility within the API. I suspect this isn't interesting enough for a news item but maybe the only way to improve visability?

wjchulme:
@evansd I don't think this change affects anything other than the outputs from `categorised_as`? If so I'll update the docstring.

evansd:
> I don't think there's a common name for this. It's really just allowing a bit more flexibility within the API. I suspect this isn't interesting enough for a news item but maybe the only way to improve visability?

Yes, I consider this a fix rather than a feature. It's just making the API work in the way it should always have worked if it wasn't for the mess we've got ourselves in with types.


> I don't think this change affects anything other than the outputs from categorised_as?

Yes, that's right. This should be a backwards compatible change simply because if you tried to use strings which looked like dates as categories before they would get mangled, so no one did.

sebbacon:
>>> Would this pattern have a well-known name to epi researchers
>>
>> I don't think there's a common name for this. It's really just allowing a bit more flexibility within the API. I suspect this isn't interesting enough for a news item but maybe the only way to improve visability?
>
> Yes, I consider this a fix rather than a feature. It's just making the API work in the way it should always have worked if it wasn't for the mess we've got ourselves in with types.

I agree this is a fix, not a feature. However, it struck me that the application of this feature to this particular problem is quite neat, and not immediately obvious.

So really, my question is more general, about the best way to label and document different kinds of strategies, in a "design patterns" kind of way

cohortextractor/tpp_backend.py
```diff
@@ -344,6 +344,16 @@ def get_case_expression(self, other_columns, column_type, category_definitions):
             raise ValueError(
                 "At least one category must be given the definition 'DEFAULT'"
             )
+
+        # If categories are date-like strings we don't want to apply the usual
+        # reformatting that we apply to dates to make sure that MSSQL
+        # interprets them consistently; instead we want them in the format we
+        # expect for outputs, which is their original ISO format. If these
```
sebbacon, line 351: Found this comment hard to understand, initially. Something like this perhaps? "We pass the `reformat_dates=False` option to the quoting function here to preserve date-like strings in the user-supplied ISO format, as opposed to converting to MSSQL-friendly format. If these...."

sebbacon (approved):

cohortextractor/tpp_backend.py
evansd: Yeah, it's a fundamentally nasty thing to try to explain. Happy to amend to whatever made it clearer for you.

elsie-h:
Thanks @evansd  - I've worked this into my study definition and variables created using 'categorised_as' are being treated as dates, which is great! However, despite specifying the dates as "YYYY-MM-DD", only the year is being stored in the variable. Could you please check that this is happening for you too and it's not just me doing something stupid?! 

Something less urgent which may or may not be related to the above is that date variables created using categorised_as are not being recognised as dates in "return_expectations".

elsie-h:
@evansd just realised it would have been helpful for me to share the study definition where I am attempting to implement this.  [This is where I define `elig_date`](https://github.com/opensafely/covid-unvaccinated-aimpaper/blob/00e7ace389c8b03fd3cb7d71b568cae0ed5c292d/analysis/study_definition.py#L70) using the updated `categorised_as`.

[Also see the lines I have commented out ](https://github.com/opensafely/covid-unvaccinated-aimpaper/blob/00e7ace389c8b03fd3cb7d71b568cae0ed5c292d/analysis/study_definition.py#L339)due to `elig_date` not being recognised as a date in `return_expectations`.

evansd:
Thanks @elsie-h, this is definitely not you doing anything stupid :smile: 

The issue with only the year being output was a bug with the dummy data generation. It would have worked OK on the real data, but that's not very helpful when you can't test your code properly. I've got a fix out for the now which should hopefully get merged and deployed soon.

The issue with not being able to write things like this
```
return_expectations={
    "date": {"earliest": "elig_date","latest": end_date},
    ...
}
```
is a more general limitation of the dummy data generation system as it currently stands. Unfortunately you just have to use fixed dates here which means that you might end up with nonsensical dummy data (e.g. events happening before they logically could have done). If that ends up preventing you from being able to test your scripts then @wjchulme might have some workarounds.

We've got plans to overhaul the dummy data generation, but it's not going to get fixed in the near future I'm afraid.

PR #659, opened 2021-09-27
clarify the meaning of with_complete_gp_consultation_history_between …

CarolineMorton:
clarify the meaning of `with_complete_gp_consultation_history_between` and `registered_with_one_practice_between`, and make minor change to remove the see above comment as the documentation is not in the order of the functions in the code. 

This works with https://github.com/opensafely/documentation/pull/425 to clarify the issue described on https://github.com/opensafely/documentation/issues/417

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -175,7 +175,11 @@ def registered_with_one_practice_between(
     return_expectations=None,
 ):
     """
-    All patients registered with the same practice through the given period
+    All patients registered with the same practice through the given period.
+
+    Note, it does not account for patients registered with one practice in a given time period, whose practice
+    changed its EHR provider during that that time period. To capture this information, please use
+    `with_complete_gp_consultation_history_between()`
 
     Args:
         start_date: start date of interest of period as a string with the format `YYYY-MM-DD`.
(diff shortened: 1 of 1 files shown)


cohortextractor/patients.py
```diff
@@ -175,7 +175,11 @@ def registered_with_one_practice_between(
     return_expectations=None,
 ):
     """
-    All patients registered with the same practice through the given period
+    All patients registered with the same practice through the given period.
+
+    Note, it does not account for patients registered with one practice in a given time period, whose practice
+    changed its EHR provider during that that time period. To capture this information, please use
+    `with_complete_gp_consultation_history_between()`
```
iaindillingham, line 182: Insofar as possible, I'd suggest using the same language in the body of the docstring as in the title; writing in the present time ("simple present tense"); and preferring non-phrasal verbs (e.g. *return* rather than *account for*). For example:

> Note, this function does not return all patients registerd with the same practice through the given time period when this practice changes its EHR provider. To capture this information, please use `with_complete_gp_consultation_history_between()`.

cohortextractor/patients.py
```diff
@@ -1528,8 +1532,12 @@ def with_complete_gp_consultation_history_between(
     return_expectations=None,
 ):
     """
+    This captures patients who are registered with one practice between the start date and end date
+    and that this practice used SystmOne continuously between those dates.
```
iaindillingham, line 1536: Consistency is important! As we're comparing two functions here, and not all the functions in this module, how about:

> All patients registered with the same practice through the given period, when the practice used SystemOne through the given period.

iaindillingham (approved):
Couple of nitpick comments, which aren't essential to address.

cohortextractor/patients.py
sebbacon: Are we sure about it only identifying SystmOne (note spelling @iaindillingham :D)?

The function name implies something different. What does it do when you run it against an EMIS backend?

iaindillingham: Oh, *that* SystmOne! 🤦🏻‍♂️

PR #662, opened 2021-09-29
Distinguishing true zeros

acagreen17:
Add in info in on using comparator to distinguish true 0s from missing values and add a link from that docstring to the Missing Values section

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -639,6 +639,8 @@ def comparator_from(
     Where a lab result is returned as e.g. `<9.5` the numeric_value component
     will contain only the value 9.5 and you will need to use this function to
     fetch the comparator into a separate column.
+    Note, the comparator can be used to distinguish true 0s from 0s which represent missing values (i.e. if the numeric_value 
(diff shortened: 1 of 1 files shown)


acagreen17:
Not sure if the link is correct or not (i.e. not sure how to link to a specific section in the documentation...)

cohortextractor/patients.py
```diff
@@ -639,6 +639,9 @@ def comparator_from(
     Where a lab result is returned as e.g. `<9.5` the numeric_value component
     will contain only the value 9.5 and you will need to use this function to
     fetch the comparator into a separate column.
+    Note, the comparator can be used to distinguish true 0s from 0s which represent missing values (i.e. if the numeric_value 
+    component is 0 but there is no value in the associated comparator column, then you can assume that the 0 is not a real value.)
```
sebbacon, line 643: Should we somehow find a way to make it clear this only applies _for an entry which is coded as something expected to have a reference range_? If you try to get reference ranges for "coded as  having long covid" it will always have a  null comparator (right @evansd?)

sebbacon (commented):
I think we need a bit more info to be really precise here...

PR #663, opened 2021-10-06
fix: Revert to using Presto client library

evansd:
This reverts the change made in #625 from using
`presto-python-client=0.7.0` to using `trino==0.306.0`. That change
passes tests but triggers 500 errors from the database in production.

Rather than revert the entire PR (which would be a revert of a revert of
a revert) we leave almost all the other changes in place and do some
import munging to access the old client under the new name. This should
buy us some time while we identify the root cause.

### DIFF
--- a/cohortextractor/trino_utils.py
+++ b/cohortextractor/trino_utils.py
@@ -7,7 +7,21 @@ from urllib.parse import unquote, urlparse
 
 import requests
 import structlog
-import trino
+
+# The Trino client ought to be close to a drop-in replacement for the Presto
+# but it isn't: despite the tests passing we get 500s from Presto in production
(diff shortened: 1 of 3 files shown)


ghickman (approved):

sebbacon (approved):
Thanks. What next - do we submit a reproducible case to EMIS, perhaps?

evansd:
I think next is a very tedious bisection process where we identify either the oldest version of `prestodb` which breaks, or the newest version of `trino` that works. Then hopefully looking at the diff will help us work out where the failure is.

Fortunately the broken version fails immediately on just a `SELECT 123;` query, so we can check quite quickly whether it exhibits the problem or not.

I don't _think_ there's an issue on the EMIS side here.

PR #669, opened 2021-10-14
fix: Fix "decision support values" to work with dynamic dates

evansd:
This involves including the date join conditions and also fully
qualifying the `patient_id` column which becomes ambiguous once we join
with the extra tables.


inglesp (approved):

StevenMaude:
We discussed the failure: it's due to this branch being based on a previous commit.

PR #676, opened 2021-11-05
fix: Allow comparator characters in category names

evansd:
Following #655 we now return "comparator" categories which have
non-alphanumeric names. In order to reference these values in
expressions like

     egfr =1  AND  egfr_comparator = '>'

we need to be able to include these values in string literals.


iaindillingham (approved):
Thanks for the clear comments 🙂

PR #680, opened 2021-11-18
fix: make upper and lower bounds consistent

HelenCEBM:
Rename example variables accordingly


CarolineMorton (approved):
Looks good

If you want this to change the documentation, once this merged, you will need to press the dependabot button on the documentation repo and it will make a new PR which will pull this change in.

PR #683, opened 2021-11-25
feat: add generator for poisson distribution in dummy data

alexwalkerepi:
This was a suggestion from the Bristol users. It gives another option in dummy data variable generation, to provide more realistic variables in some cases.

The advantage of poisson distributions over e.g. a normal distribution is that it fits much better when for example counting events. With a normal distribution, you're likely to end up with negative numbers, which is nonsensical when you're counting events.

I think the code additions themselves are straightforward.

### DIFF
--- a/cohortextractor/expectation_generators.py
+++ b/cohortextractor/expectation_generators.py
@@ -3,7 +3,7 @@ from datetime import datetime
 
 import numpy as np
 import pandas as pd
-from scipy.stats import expon, norm, rv_discrete, uniform
+from scipy.stats import expon, norm, rv_discrete, uniform, poisson
 
 
 def generate_ages(population, max_age=110):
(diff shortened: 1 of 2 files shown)


evansd (approved):
This is great, thanks @alexwalkerepi!

The formatting check was complaining about the order of the imports from `scipy.stats`. That may sound incredibly pedantic but there's a tool called [isort](https://pycqa.github.io/isort/) which automatically reformats all your imports into sensibly ordered, coherent groups and the check is just making sure that that's be run.

alexwalkerepi:
I'm not sure how the order it prefers is any more logical than the original, but as it has an opinion and I don't, sure :)

evansd:
Oh the order is just alphabetical. The grouping is to do with whether packages are standard library, third-party or your own application code. But it isn't so much that the choices it makes are especially great in themselves. The lovely thing is just never having to think about this stuff again: you just stick the import wherever you want and then run `isort` (which you can configure your editor to do automatically if you want). It's the same with `black` which reformats all your code so you never have to think about how where you're wrapping a line or how your indenting a dictionary or what sort of quotes your using.

PR #687, opened 2021-11-30
Fix dummy data validation for date categories

### LINKED ISSUE #686: Dummy data files with date columns
It's now possible to pass a dummy data file when generating a cohort, avoiding some clunkiness associated with [the expectations framework][1].

opensafely/documentation#499 describes how a researcher passed a dummy data file with a date column. However, the corresponding column in the study definition was the result of calling `patients.categorised_as`, meaning that if the dummy data file hadn't been passed, then the date column would have been a string column.

When verifying dummy data files, cohort-extractor assumes that a date column (`query_args["column_type"] == "date"`) will be associated with a `date_format` argument. This assumption doesn't hold, here, so cohort-extractor raised the following error:

```
Traceback (most recent call last):
  File "/usr/local/bin/cohortextractor", line 33, in <module>
    sys.exit(load_entry_point('opensafely-cohort-extractor', 'console_scripts', 'cohortextractor')())
  File "/app/cohortextractor/cohortextractor.py", line 668, in main
    generate_cohort(
  File "/app/cohortextractor/cohortextractor.py", line 151, in generate_cohort
    _generate_cohort(
  File "/app/cohortextractor/cohortextractor.py", line 202, in _generate_cohort
    study.to_file(
  File "/app/cohortextractor/study_definition.py", line 92, in to_file
    validate_dummy_data(self.covariate_definitions, dummy_data_file)
  File "/app/cohortextractor/validate_dummy_data.py", line 31, in validate_dummy_data
    validator = get_binary_validator(col_name, query_args)
  File "/app/cohortextractor/validate_dummy_data.py", line 174, in get_binary_validator
    if query_args["date_format"] == "YYYY":
KeyError: 'date_format'
```

Here's the line, with some context:

https://github.com/opensafely-core/cohort-extractor/blob/1514a7abbeda66157345ff18e69107c694c5b267/cohortextractor/validate_dummy_data.py#L172-L175

The line and the `KeyError` aren't especially informative. Instead, we could...

* return `date_validator_day` (we might need to replace the call to `isinstance` with a call to [`is_datetime64_any_dtype`][2])
* raise a more informative error
* something else 🤔 

[1]: https://docs.opensafely.org/study-def-expectations/
[2]: https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_datetime64_any_dtype.html

evansd:
Closes #686

### DIFF
--- a/cohortextractor/emis_backend.py
+++ b/cohortextractor/emis_backend.py
@@ -270,7 +270,9 @@ class EMISBackend:
         else:
             raise ValueError(f"Unhandled column type: {column_type}")
 
-    def get_case_expression(self, other_columns, column_type, category_definitions):
+    def get_case_expression(
+        self, other_columns, column_type, category_definitions, date_format=None
+    ):
         category_definitions = category_definitions.copy()
         defaults = [k for (k, v) in category_definitions.items() if v == "DEFAULT"]
         if len(defaults) > 1:
--- a/cohortextractor/process_covariate_definitions.py
+++ b/cohortextractor/process_covariate_definitions.py
@@ -20,6 +20,7 @@ def process_covariate_definitions(covariate_definitions):
         covariate_definitions
     )
     covariate_definitions = add_column_types(covariate_definitions)
+    covariate_definitions = set_format_for_date_categories(covariate_definitions)
     check_for_consistent_aggregate_date_formats(covariate_definitions)
     return covariate_definitions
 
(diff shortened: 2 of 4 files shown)


ghickman (approved):

iaindillingham:
I made a couple of comments on [Slack][1] that I should have made here. Essentially, by ensuring that `query_args["date_format"]` is set to `"YYYY-MM-DD"`, `get_binary_validator` will return `date_validator_day` as the validator. I wasn't sure whether this validator would validate the researcher's dummy data.

It does!

My suggestion to use [`is_datetime64_any_dtype`][2] was also incorrect; I hadn't appreciated that we validate each value in the column, rather than the column itself.

[1]: https://ebmdatalab.slack.com/archives/C01SE1WB6AW/p1638305112290700
[2]: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.api.types.is_datetime64_any_dtype.html

evansd:
@iaindillingham Yes, the basic problem is that in #656 I introduced a new way of creating date type columns but then failed to think through all the consequences of that -- hence I felt some responsibility to fix this!

iaindillingham:
You shouldn't have 🙂 It was a great opportunity to sharpen my code comprehension skills.

evansd:
Well you managed to isolate the problem which is 90% of the work in this kind of situation.

PR #705, opened 2021-12-23
feat: switch to using pymssql as a db driver

### LINKED ISSUE #704: Switch from pyODBC and cTDS to pymssql
At the moment we use [pyODBC](https://pypi.org/project/pyodbc/) for test setup (because it's supported by SQLAlchemy) and [cTDS](https://pypi.org/project/ctds/) for querying the database.

We used to use `pyODBC` for everything but we started to get errors like the below when downloading more that about 2GB of data:
```
pyodbc.OperationalError: (
  '08S01',
  '[08S01] [Microsoft][ODBC Driver 17 for SQL Server]TCP Provider: Error code 0x2746 (10054) (SQLGetData)'
)
```

We switched to using `cTDS` in [this PR](https://github.com/opensafely-core/cohort-extractor/pull/286). This doesn't come with pre-built wheels, and the installation process is a bit fiddly, so we built our own Linux binary wheel here:
https://github.com/opensafely-core/ctds-binary

At the time we also evaluated [pymssql](https://github.com/pymssql/pymssql) this has a much simpler installation story (binary wheels available for many platforms), seems to have a number of active maintainers and is [praised](https://docs.sqlalchemy.org/en/13/dialects/mssql.html#module-sqlalchemy.dialects.mssql.pymssql) in the SQLAlchemy docs. But we hit [this](https://github.com/pymssql/pymssql/issues/500) alarming "corrupted unsorted chunks" error when our results hit a certain size, and so gave up.

Eventually though we started getting increasingly regular "EOF from server" errors when downloading large result sets, even with `cTDS`. So we ended up taking a completely different "batched download" approach here: https://github.com/opensafely-core/cohort-extractor/pull/322

This means that we should no longer hit the memory issue with `pymssql` and there are a number of advantages to using it instead:
 * We don't need to rely on the Microsoft apt repository to install the dependencies of `pyODBC`, which has proved unreliable in the past.
 * We don't need to build and maintain our own wheels as we do for `cTDS`.
 * We can use the same library for test setup and for querying (`pymssql` has SQLAlchemy support).
 * We're already using `pymssql` in [databuilder](https://github.com/opensafely-core/databuilder).

bloodearnest:
Fixes #704

Tests and prod now use the same driver, and installation story is
easier.

The only significant changes were serialising enums values, and explicit
transaction management.


evansd (approved):
Thanks for getting this working. It looks like it was the opposite of fun :smile: 

I think there's just one small tweak which needs doing and that's adding `pymssql` to the list here:
https://github.com/opensafely-core/cohort-extractor/blob/b9ed15deffdced21a0180d7b8118d01657955448/tests/test_study_definition.py#L241

Hopefully the comments in that test explain enough of the context for this.

PR #707, opened 2022-01-04
ci: Fix broken Trino version check

StevenMaude:
When the separate requirements files were added in #605, this broke
because the possibility of multiple files existing wasn't accounted for.

`grep` helpfully has `--no-filename` by default when only one file is
searched. Now we're searching two requirements files, so that implicit
setting broke.

The full flags to `grep` are now included, because it's not too obvious
that `-h` means `--no-filename`.


StevenMaude:
Working `workflow_dispatch` run: https://github.com/opensafely-core/cohort-extractor/actions/runs/1654499720

iaindillingham (approved):

Issue #708, opened 2022-01-06
Update link in high cost drugs documentation


rose-higgins:
Update the link in the high cost drugs variable documents from 
https://github.com/opensafely/highcostdrugs-research/blob/master/analysis/High%20Cost%20Drugs%20Dataset.ipynb

to

https://wellcomeopenresearch.org/articles/6-360

PR #709, opened 2022-01-06
docs: Fix link in high cost drugs description

rose-higgins:
Issue for this fix is:

#708

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -2244,8 +2244,8 @@ def with_high_cost_drugs(
     """
     Returns data from the High Cost Drugs Dataset
 
-    More details available in this notebook:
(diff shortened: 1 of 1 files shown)


iaindillingham (approved):
To merge this PR, you'll need to turn its title into a [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) by adding a prefix. `docs:` seems the most appropriate, to give `docs: Fix link in high cost drugs description`.

evansd (approved):
Thanks @rose-higgins! The Semantic Pull Request check is failing because it isn't very good, not because there's anything wrong with your PR so I'm happy to merge as is.

There's a separate step we'll have to do to update the documentation site (the whole documentation pipeline is a bit janky at the moment) so your fix may not appear in the real world for a little while, but it will do next time we produce a new release of cohortextractor.

PR #712, opened 2022-01-11
ci: fix publish action to not have single quotes in VERSION

bloodearnest:


.github/workflows/build_and_publish.yaml
```diff
@@ -75,7 +75,7 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v1
     - name: set VERSION environment variable
-      run: echo "VERSION='${{github.event.inputs.tag || needs.tag-new-version.outputs.tag}}'" >> $GITHUB_ENV
+      run: echo "VERSION=v${{github.event.inputs.tag || needs.tag-new-version.outputs.tag}}" >> $GITHUB_ENV
```
ghickman, line 78: Does `v` do anything here? Or is it just the literal string `v`?

ghickman (approved):

.github/workflows/build_and_publish.yaml
bloodearnest: Just a literal v as that's been our convention for tag names

PR #714, opened 2022-01-13
fix: clamp dates to pandas max

bloodearnest:
TPP uses the date `9999-12-31 00:00:00` to represent a max future date,
but this is too large for `pandas.Timestamp` when we download the
results.

So, we clamp this specific date to `pandas.Timestamp.max`.

I chose to be conservative about what dates we clamp, but in theory we
could clamp anything over `2262-04-11 23:47:16.854775807`. But that
would mean actually parsing the date string, which is what pandas is
supposed to do for us, and python stdlib does not have a generic parser
for. And I wasn't sure about the semantics.

### DIFF
--- a/cohortextractor/pandas_utils.py
+++ b/cohortextractor/pandas_utils.py
@@ -138,4 +138,20 @@ def memoize(fn):
     return cache().__getitem__
 
 
-to_datetime = memoize(pandas.to_datetime)
+# dates that different systems use to represent the max date, but are greater
+# than pandas.Timestamp can represent, so we clamp to pandas.Timestamp.max
+MAX_DATES = [
+    "9999-12-30 00:00:00",  # TPP
+]
+
+
+@memoize
+def to_datetime(datestr):
+    try:
+        return pandas.to_datetime(datestr)
+    except pandas.errors.OutOfBoundsDatetime:
+        if datestr in MAX_DATES:
+            # clamp to pandas max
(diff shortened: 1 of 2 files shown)


cohortextractor/pandas_utils.py
```diff
@@ -138,4 +138,20 @@ def __missing__(self, key):
...
+@memoize
+def to_datetime(datestr):
+    try:
+        return pandas.to_datetime(datestr)
+    except pandas.errors.OutOfBoundsDatetime:
+        if datestr in MAX_DATES:
+            # clamp to pandas max
+            return pandas.Timestamp.max
+        else:
+            raise
```
evansd, line 157: Good idea to do this!

evansd (approved):
:+1 This looks sensible to me

bloodearnest:
One thing that occurred to me post PR:

Will this only affect feather files? Or csvs too? I guess I ideally we'd have the same clamping behaviour across all formats?

evansd:
This will affect all binary formats (so Stata and R, as well as feather) but not CSVs. We've painted ourselves into a bit of a corner here as there will no doubt be existing studies that rely on the 9999 date so we can't change it easily. For binary formats it was just broken before so we know there can't be any code relying on that behaviour.

I think we're OK tolerating this kind of inconsistency here because we can address this, and other issues like NULL handling, in the databuilder.

Issue #716, opened 2022-01-19
Investigate HEST_OS_study errors


rebkwok:
The `covid_admission_primary_diagnosis` variable in this (study definition)[https://github.com/opensafely/HEST_OS_study/blob/main/analysis/study_definition.py] is sometimes returning no data which causes (this)[https://github.com/opensafely/HEST_OS_study/blob/main/analysis/full_table_script.R#L91] to fail
```
covid_admission_primary_diagnosis=patients.admitted_to_hospital(
        returning="primary_diagnosis",
        with_these_diagnoses=covid_codelist,
        on_or_after= index_date,
        find_first_match_in_period=True,  
        date_format="YYYY-MM-DD", 
        return_expectations={"date": {"earliest": index_date},"incidence" : 0.25,
            "category": {"ratios": {"U071":0.5, "U072":0.5}},
        },
    ),```

The last run on TPP failed https://github.com/opensafely/HEST_OS_study/blob/main/analysis/full_table_script.R#L91

(Note the use of the `date_format` kwarg when the returning value is a string, not a date - this appears to be a red herring;  the date_format kwarg will be ignored)

rebkwok:
I've tried running the SQL that is generated from the study definition for the specified population and covid_admission_primary_diagnosis only.  It produces results, but all the covid_admission_primary_diagnosis column is empty, and there is an error message
The invalid escape character "\\" was specified in a LIKE predicate
This is coming from the covid_admission_primary_diagnosis variable, which generates:
-- Query for covid_admission_primary_diagnosis
SELECT * INTO #covid_admission_primary_diagnosis FROM (
    SELECT
    t.Patient_ID AS patient_id,
    t.primary_diagnosis AS primary_diagnosis
    FROM (
        SELECT APCS.Patient_ID, Spell_Primary_Diagnosis AS primary_diagnosis,
        ROW_NUMBER() OVER (
            PARTITION BY APCS.Patient_ID
            ORDER BY Admission_Date ASC, APCS.APCS_Ident
            ) AS rownum
            FROM APCS
            INNER JOIN APCS_Der
            ON APCS.APCS_Ident = APCS_Der.APCS_Ident
    WHERE Admission_Date >= '20210514' AND (Der_Diagnosis_All LIKE '%[^A-Za-z0-9]U071%' ESCAPE '\\' OR Der_Diagnosis_All LIKE '%[^A-Za-z0-9]U072%' ESCAPE '\\')
    ) t
[...]

PR #717, opened 2022-01-19
Change escape character used in TPP backend

rebkwok:
Using `//` as an escape character seems to be problematic - see #716.  Replace it with `!` instead


ghickman (approved):

tests/test_tpp_backend.py
```diff
@@ -2047,22 +2047,22 @@ def test_quote_against_db():
 
 def test_escape_like_query_fragment():
     assert escape_like_query_fragment("foo") == "foo"
-    assert escape_like_query_fragment("foo%bar_") == "foo\\%bar\\_"
+    assert escape_like_query_fragment("foo%bar_") == "foo!%bar!_"
 
 
 def test_escape_like_query_fragment_against_db():
```
evansd, line 2053: Oh, I did test this when I first wrote it! Now I'm very confused as to what was causing the error we saw earlier. But in any case, I'm pretty sure this change will fix it because we don't need to worry about double-escaping.

evansd (approved):

tests/test_tpp_backend.py
evansd: I think I get why the tests passed because in the context of a Python string `\\` is how you spell `\` (i.e. `print("\\")` outputs `\`) but I don't understand how the double blackslash then made its way into SQL and triggered that error. Anyway, thanks for fixing!

PR #721, opened 2022-01-22
fix: datetime clamping work with actual string of "9999-12-31"

bloodearnest:


evansd (approved):

cohortextractor/pandas_utils.py
```diff
@@ -162,7 +162,9 @@ def wrapper(key):
 # dates that different systems use to represent the max date, but are greater
 # than pandas.Timestamp can represent, so we clamp to pandas.Timestamp.max
 MAX_DATES = [
-    "9999-12-30 00:00:00",  # TPP
+    # TPP
+    "9999-12-30",
```
rebkwok, line 166: Is the ceiling date not "9999-12-31"?

evansd: That is a very good point

bloodearnest: Indeed! Fixed here: https://github.com/opensafely-core/cohort-extractor/pull/722

PR #723, opened 2022-01-24
Add date expressions for first and last day of NHS financial year

### LINKED ISSUE #719: New helper function: `first/last_day_of_nhs_financial_year`
The standard NHS financial (or reporting) year runs 1st April - 31st March. This is for example used for calculating practice achievement towards QOF targets.

Example use case: run a report on QOF measures `by month` (to allow monthly tracking of progress) which can easily use the last day of the relevant financial year as the "end date"

This should be very similar to our existing functions `first/last_day_of_year`

rebkwok:
Fixes #719

### DIFF
--- a/tests/test_tpp_backend.py
+++ b/tests/test_tpp_backend.py
@@ -3761,6 +3761,83 @@ def test_dynamic_index_dates_with_invalid_expression():
         )
 
 
+@pytest.mark.parametrize(
+    "index_date,expected_patient_ids",
+    [
+        ("2020-02-15", ["2", "3"]),  # nhs financial year 2019-04-01 to 2020-03-31
+        ("2017-05-15", ["4", "5"]),  # nhs financial year 2017-04-01 to 2018-03-31
+    ],
+)
+def test_nhs_financial_year_date_expressions(index_date, expected_patient_ids):
+    session = make_session()
+    session.add_all(
+        [
+            # Event too early
+            Patient(
+                Patient_ID=1,
+                DateOfBirth="1980-01-01",
+                CodedEvents=[CodedEvent(ConsultationDate="2012-12-15", CTV3Code="foo")],
+            ),
+            # Events in range 2019-04-01 to 2020-03-31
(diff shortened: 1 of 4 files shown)


tests/test_tpp_backend.py
```diff
@@ -3761,6 +3761,83 @@ def test_dynamic_index_dates_with_invalid_expression():
...
+                Patient_ID=9,
+                DateOfBirth="1980-01-01",
+                CodedEvents=[CodedEvent(ConsultationDate="2020-04-01", CTV3Code="foo")],
+            ),
+        ]
+    )
+    session.commit()
+    study = StudyDefinition(
+        index_date=index_date,
+        population=patients.with_these_clinical_events(
```
iaindillingham, line 3829: When you tested against the EMIS backend, you tested the expressions with the context of `population` and `event_date`; the former because it's special and the latter because it's not, I assume. Why not do the same here?

I'm sure there's a good reason 🙂

iaindillingham (approved):
Very clear! I feel I've learnt a lot more about cohort-extractor, just from looking at this PR.

tests/test_tpp_backend.py
rebkwok: I did the tpp tests first, and it made sense to just test against the patient ids that were returned.  `population` on its own just gets you a set of `patient_id` in the results.  I tried to do the same thing in the emis tests, but the emis backend generates patient ids (I confess I didn't look too closely at exactly what it does), so the patient ids in the results were always 0 and 1.  The emis one is only that way to have some way to checking that the 2 patients it returns are the expected ones.

iaindillingham: Ah, of course! Yes, the EMIS backend handles patient IDs differently, making is much harder to test against just `population`. More here:

https://github.com/opensafely-core/cohort-extractor/blob/84dddec21177cc45882569bcf52cbf9635ca6c57/cohortextractor/emis_backend.py#L1585-L1596

PR #725, opened 2022-01-25
fix: Set expectations for value from/in functions

iaindillingham:
`StudyDefinition.make_df_from_expectations` expects each function in the `patients` module to accept a `return_expectations` argument.

https://github.com/opensafely-core/cohort-extractor/blob/077dbd8819eb5c5f2ca9101dad2c5e1350b0e9d2/cohortextractor/study_definition.py#L294

Previously, `patients.with_value_from_file` and `patients.which_exist_in_file` didn't accept these arguments. However, these functions weren't called from a smoke-test, so @iaindillingham didn't notice that these arguments were missing. Here, we add these arguments and add a smoke-test.

I considered rewriting the above as:

```python
return_expectations = definition_args.get("return_expectations") or {}
```

However, I feel it's safer to make `with_value_from_file` and `which_exist_in_file` consistent with the other functions in the `patients` module.

Thanks to @rriefu for raising this issue, following [this test](https://github.com/opensafely/amr-uom-brit/runs/4925616820).

Something to note is that `with_value_from_file` and `which_exist_in_file` shouldn't need return expectations; the contents of the files that are passed _are_ the return expectations. However, unpicking cohort-extractor to make this so seems excessive. I'll revisit in a separate issue, if this starts to trip people up.

### DIFF
--- a/cohortextractor/study_definition.py
+++ b/cohortextractor/study_definition.py
@@ -68,11 +68,12 @@ class StudyDefinition:
     ):
         if expectations_population:
             df = self.make_df_from_expectations(expectations_population)
-            # Add a patient ID - a randomly generated integer from an
-            # array 10x larger than the cohort.
-            df["patient_id"] = default_rng().choice(
-                (len(df) * 10), size=len(df), replace=False
-            )
+            if "patient_id" not in df:
+                # Add a patient ID - a randomly generated integer from an
+                # array 10x larger than the cohort.
+                df["patient_id"] = default_rng().choice(
+                    (len(df) * 10), size=len(df), replace=False
+                )
             if not is_csv_filename(filename):
                 # This slightly convoluted approach means that dummy data
                 # intended for a binary output format gets passed through the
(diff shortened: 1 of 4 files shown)


evansd:
Sorry, just looking at this now. I think an alternative approach would be to add these functions to the exception condition here:
https://github.com/opensafely-core/cohort-extractor/blob/077dbd8819eb5c5f2ca9101dad2c5e1350b0e9d2/cohortextractor/study_definition.py#L287-L288

cohortextractor/patients.py
```diff
@@ -2443,19 +2449,17 @@ def with_value_from_file(f_path, returning, returning_type, date_format="YYYY-MM
             * `YYYY-MM-DD` The default value.
             * `YYYY-MM`
             * `YYYY`
-
-    This function does not accept a `return_expectations` argument because the file can contain dummy data.
+        return_expectations: as described elsewhere.
```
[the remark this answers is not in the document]
iaindillingham: As written, return expectations are necessary, either through the `return_expectations` argument or the `default_expectations` property, and they replace the data in the file. Both characteristics came as a surprise to me; the first motivated this PR; I discovered the second when writing the smoke test. I'll investigate Dave's alternative approach and, if successful, rework the documentation. Thanks for taking a look 🙂

evansd:
Thinking about this a bit more, adding it to the list of excluded functions like `aggregate_of` isn't quite sufficient. Because we still need to do _something_ with those functions when generating dummy data. In this case, I think that something is: actually load the supplied file and pull the referenced columns out.

You can see where calls to the `aggregate_of` method are handled here:
https://github.com/opensafely-core/cohort-extractor/blob/077dbd8819eb5c5f2ca9101dad2c5e1350b0e9d2/cohortextractor/study_definition.py#L355-L379

So I suppose it would be a case of adding another loop over the code to handle the `with_value_from_file/which_exist_in_file` methods.

If it makes you feel any better, I _always_ forget about dummy data when adding new functions. As an example, [here](https://github.com/opensafely-core/cohort-extractor/pull/148) is the PR where I added the minimum_of/maximum_of functions. And [here](https://github.com/opensafely-core/cohort-extractor/pull/638) is the PR where Jon added dummy data support for it, making it actually usable by researchers.

Note the 14 month time interval!

iaindillingham:
Thanks for your comments @rebkwok and @evansd. I've taken Dave's alternative approach; neither `with_value_from_file` nor `which_exist_in_file` accept `return_expectations`. In each case, the expectations are the data within, or inferred from, the file; this is consistent with the documentation and, hence, the intended behaviour.

I've added some checks to `make_df_from_expectations` that mirror those within the TPP backend (`TPPBackend.patients_with_value_from_file`).

cohortextractor/study_definition.py
```diff
@@ -280,11 +281,49 @@ def add_day_to_date(val):
...
+
+            returning = definition_args.get("returning")
+            returning_type = definition_args.get("returning_type")
+            if returning is not None and returning_type is None:
+                raise ValueError(f"No `returning_type` defined for {colname}")
+            if returning is None and returning_type is not None:
+                raise ValueError(f"No `returning` defined for {colname}")
+            if returning is not None and not re.fullmatch(r"\w+", returning, re.ASCII):
+                raise ValueError(
+                    f"{colname} should contain only alphanumeric characters and the underscore character"
```
rebkwok, line 303: These checks duplicate what the actual functions do; is it worth pulling them out to a shared utility somewhere?

cohortextractor/study_definition.py
```diff
@@ -313,7 +352,11 @@ def make_df_from_expectations(self, population):
         # its incidence calculated as a mask
         for colname, dtype in self.pandas_csv_args["dtype"].items():
             definition_args = self.pandas_csv_args["args"][colname]
-            if definition_args["funcname"] == "aggregate_of":
+            if definition_args["funcname"] in [
+                "aggregate_of",
+                "with_value_from_file",
+                "which_exist_in_file",
```
rebkwok, line 358: very minor nitpick; I'd make this list a variable so we don't have to repeat the actual function names.

rebkwok (approved):
LGTM. A couple of minor suggestions on repeated code.

cohortextractor/study_definition.py
iaindillingham: I wasn't sure, to be honest. In the actual functions, I've raised `TypeError`; within the `study_definition` module, `ValueError` is the norm. (I mentioned being unsure about what type of error to raise during today's meeting.) In a couple of cases, the actual functions don't have access to the column names, so the error messages can't be as specific (and hence, useful). For these reasons, I didn't pull the checks into a shared utility.

cohortextractor/study_definition.py
iaindillingham: I wasn't sure about this, either 🙂 Specifically, I wasn't sure how to distinguish `aggregate_of` from `with_value_from_file` and `which_exist_in_file`; we always use the last two, together; and sometimes use all three, together. So, how to name these variables? I couldn't quickly think of good names, so repeated the actual function names.

We could call them according to the types of function they represent, like this:

```python
aggregation_functions = ["aggregate_of"]
user_file_functions = ["with_value_from_file", "which_exist_in_file"]
```

But these names don't really communicate why we're handling these types of function differently.

PR #726, opened 2022-01-26
Summary (mean, min, max) recorded values across a range of dates

### LINKED ISSUE #724: Mean/min/max methods for recorded value over a time period
Currently we have `patients.mean_recorded_value`, which takes a date range and returns the mean recorded value on the most recent day of measurement. I'm proposing that we extend this to:
1) Allow calculation of a mean value of all recordings during the date range, not solely from the most recent day
2) Provide equivaluent min/max functions (`patients.min_recorded_value` and `patients.max_recorded_value`

### Rationale:
1) It's straightforward
The code for `mean_recorded_value` can be quite easily modified to calculate the mean for the entire date range.  Calculating min/max is almost identical, and should just be a matter of swapping out an SQL function.

2) It's useful to an external user
And to internal users - @HelenCEBM reports having used a workaround for finding max values by creating a series of variables for throughout the period using the `maximum_of` function.

3) The current method name and documentation is misleading
A `mean_recorded_value` which takes a date range implies that it's going to return the mean within that date range.  
The current docs say this method will "Return patients' mean recorded value of a numerical value as defined by a codelist on a particular day within the defined period." and describe the `on_most_recent_day_of_measurement` parameter as a "boolean flag for requesting measurements be on most recent date".
Since you can't specify the specific date to return (except by providing the same date as the start and end of the `between` parameter), this implies that `on_most_recent_day_of_measurement=False` will give you the mean of the whole range (in fact, it will raise an `AssertionError` because the method currently only supports `on_most_recent_day_of_measurement=True`)

4) It should be available for free with the new databuilder
Data builder will allow more generic filtering and aggregating on tables, so it should eventually be possible to do something like
```
max_value = clinical_events.filter("2020-02-01" < clinical_events.date <= "2020-03-01").max_for_patient(clinical_events.value).select_column(clinical_events.value)
```
So adding this functionallity to cohort-extractor won't add new special things that need to be implemented specially in data builder too

rebkwok:
Fixes #724 
See issue for rationale for adding this functionality.

Extends `mean_recorded_value` to optionally cover the entire provided range, not just the most recent day of measurement, and adds the equivalent methods for min_recorded_value` and `max_recorded_value`

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -366,7 +366,8 @@ def mean_recorded_value(
             Filters results to measurements between the two dates provided (inclusive).
             The two dates must be in chronological order.
         include_measurement_date: a boolean indicating if an extra column, named `<variable_name>_date_measured`,
-            should be included in the output.  Can only be used if include_measurement_date is `True`.
+            should be included in the output.  This option can only be True when `on_most_recent_day_of_measurement`
+            is `True` (i.e. the value returned is the mean of measurements on a single day).
         date_format: a string detailing the format of the dates to be returned. It can be `YYYY-MM-DD`,
             `YYYY-MM` or `YYYY` and wherever possible the least disclosive data should be returned. i.e returning
             only year is less disclosive than a date with day, month and year. Only used if
@@ -384,7 +385,7 @@ def mean_recorded_value(
         This creates a variable `bp_sys` returning a float of the most recent systolic blood pressure from
         the record within the time period. In the event of repeated measurements on the same day, these
         are averaged. Patient who do not have this information
-        available do not return a value.  The date of measurement is returned as `bp_sys_date_measured`:
+        available do not return a value.  The date of measurement is returned as `bp_sys_date_measured`, in YYYY-MM format:
 
             bp_sys=patients.mean_recorded_value(
                 systolic_blood_pressure_codes,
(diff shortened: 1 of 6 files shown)


evansd (approved):
I won't claim to have gone through this with a fine-toothed comb, but it does look correct to me and I can see it's being tested. It might benefit from someone else checking they're happy with the detail though.

Incidentally, it's great to have more people adding meaty features to cohortextractor v1 because I'm confident it will result it more people who are absolutely convinced we needed to do the databuilder rewrite :smile:

rebkwok:
> Incidentally, it's great to have more people adding meaty features to cohortextractor v1 because I'm confident it will result it more people who are absolutely convinced we needed to do the databuilder rewrite smile

Absolutely :).  In databuilder I'm pretty sure we'd just get this for free (when the aggregation methods are available).

cohortextractor/patients.py
```diff
@@ -383,7 +384,7 @@ def mean_recorded_value(
         This creates a variable `bp_sys` returning a float of the most recent systolic blood pressure from
         the record within the time period. In the event of repeated measurements on the same day, these
         are averaged. Patient who do not have this information
-        available do not return a value:
+        available do not return a value.  The date of measurement is returned as `bp_sys_date_measured`:
```
iaindillingham, line 387: We may also wish to note that the date of measurement is to year-month resolution.

cohortextractor/patients.py
```diff
@@ -397,11 +398,197 @@ def mean_recorded_value(
...
+        creatinine=patients.mean_recorded_value(
+            creatinine_codes,
+            on_most_recent_day_of_measurement=False,
+            between=["2019-09-16", "2020-03-15"],
+            return_expectations={
+                "float": {"distribution": "normal", "mean": 150, "stddev": 200},
+                "date": {"earliest": "2019-09-16", "latest": "2020-03-15"},
+                "incidence": 0.75,
+            },
+        ),
```
iaindillingham, line 414: If a researcher were to add `include_measurement_date=True` and `date_format="YYYY-MM-DD`, what date would be returned? (Similarly for month and year, especially when the range spans months and years.)

cohortextractor/tpp_backend.py
```diff
@@ -770,48 +770,114 @@ def patients_most_recent_bmi(
...
+            {coded_event_table}.Patient_ID = days.Patient_ID
+            AND {coded_event_table}.{coded_event_column} IN ({codelist_sql})
+            AND CAST({coded_event_table}.ConsultationDate AS date) = days.date_measured
+            )
+            GROUP BY days.Patient_ID, days.date_measured
+            """
+        else:
+            assert (
+                include_date_of_match is False
+            ), "Can only include measurement date if on_most_recent_day_of_measurement is True"
```
iaindillingham, line 816: Answers my earlier question 🙂

iaindillingham (approved):
I've left a few documentation-related nitpick comments.

cohortextractor/patients.py
rebkwok: In this example, because `on_most_recent_day_of_measurement=False`, it would raise an exception (an AssertionError), because it doesn't make any sense to inclde the measurement date if the mean is taken over more than one day.  I didn't add that to the docstring because there are various other places where cohortextractor just asserts that invalid kwargs aren't included, without explicitly documenting it.  It'll fail immediately on a local run, with a message, so it should be obvious what's wrong.
 
If `on_most_recent_day_of_measurement=True` and `include_measurement_date=True`, it'll return the single date that was the most recent day of measurement.  If the date format was `YYYY`, it'll just return that day with year, which won't necessarily be any use to anyone if the range is a number of months within the same year, but I guess that's up to the individual researcher.

There's also another way to get the date.  If you leave out the i`nclude_measurement_date=True` (or make it False), you can define another variable to get the date, by doing `patients.date_of(creatinine, date_format="YYYY=MM-DD")`  (as we do in the tests), which also lets you give it a custom name.  I'm guessing that this is the more recent way to do it, but I followed the documentation of the existing `mean_recorded_value`.  Maybe it makes sense to include that option in the docstrings too.

PR #727, opened 2022-01-28
fix: Fix population size for dummy data

iaindillingham:
When we generate dummy data, `with_value_from_file` and `which_exist_in_file` in effect modify the population size from the value given within *project.yaml* to the number of rows in the user-supplied CSV file. However, when a variable in a study definition includes a dependant date, then an `ndarry` of length "population size" is assigned to a data frame of length "number of rows", which raises a `ValueError`.

Here, we add add such a variable to a study definition to recreate the error (thanks for reporting it, @rriefu). We then ensure that when these functions are present in a study definition, the population size is set to the number of rows in the user-supplied CSV file.

---

Pandas indexing gotcha:

```python
df_100 = pandas.DataFrame(data=range(100), columns=["col_a"])
df_200 = pandas.DataFrame(data=range(200), columns=["col_b"])
df_100["col_b"] = df_200["col_b"]  # inner-join
len(df_100)  # 100

df_100 = pandas.DataFrame(data=range(100), columns=["col_a"])
df_100["col_b"] = numpy.arange(200)  # Or any function that returns an ndarray
# ValueError: Length of values (200) does not match length of index (100)
```

### DIFF
--- a/cohortextractor/study_definition.py
+++ b/cohortextractor/study_definition.py
@@ -6,6 +6,7 @@ import shutil
 from pathlib import Path
 
 import pandas as pd
+import structlog
 from numpy.random import default_rng
 
 from .csv_utils import is_csv_filename
@@ -19,6 +20,8 @@ from .expectation_generators import generate
 from .pandas_utils import dataframe_from_rows, dataframe_to_file, dataframe_to_rows
 from .process_covariate_definitions import process_covariate_definitions
 from .validate_dummy_data import validate_dummy_data
+
+logger = structlog.get_logger()
 
 
 class StudyDefinition:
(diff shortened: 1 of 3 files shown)


cohortextractor/study_definition.py
```diff
@@ -314,6 +314,10 @@ def make_df_from_expectations(self, population):
 
             df["patient_id"] = user_df["patient_id"]
             df[colname] = user_df[returning] if returning is not None else 1
+            # The population should be the same size as the number of patients in the
+            # user-supplied CSV file. This ensures we generate no more dummy data than
+            # is necessary.
+            population = len(df)
```
rebkwok, line 320: Would it be useful to log this, to provide info to the user if they tried to set an expectation_population that was different to the number of patients in the CSV file?  It might not be immediately obvious why changing the expectation_population doesn't have an effect in this case

rebkwok (approved):
One suggestion for logging info for the user, if you think it's helpful

cohortextractor/study_definition.py
iaindillingham: It would! Great suggestion. I've logged this in fabc874.

rebkwok (approved):

PR #729, opened 2022-02-02
Add Therapeutics dataset (Antibody and antivirals)

### LINKED ISSUE #713: Antibody and antiviral deployment

rebkwok:
Fixes #713 

Adds the new Therapeutics dataset with a `with_covid_therapeutics` study definition query function, as described in #713 

```
covid_therapeutics = patients.with_covid_therapeutics(
	returning = "date",
	with_these_statuses = status_codelist,
	with_these_therapeutics = therapeutic_codelist,
	with_these_indications = indication_codelist,
	on_or_after = "index_date",
	date_format ="YYYY-MM-DD",
        return_first_match_in_period = True,
)
```
Note that there are some duplicate rows for patients, which aer probably not valid, and some "bad" dates that are probably typos (e.g. year 5202).   I'm not excluding or attempting to fix any of these - for now Iwell leave that to the researcher to exclude by setting a sensible date range to filter on, and by creating variables to target specific therapeutics of interest

### DIFF
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -29,6 +29,7 @@ class TPPBackend:
         self.covariate_definitions = covariate_definitions
         self.temporary_database = temporary_database
         self.next_temp_table_id = 1
+        self._therapeutics_table_name = None
         self.queries = self.get_queries(self.covariate_definitions)
 
     def to_file(self, filename):
@@ -2787,6 +2788,31 @@ class TPPBackend:
               OPA.Patient_ID
             """
 
+    def create_therapeutics_table(self):
+        if self._therapeutics_table_name is None:
+            self._therapeutics_table_name = self.get_temp_table_name("therapeutics")
+            collation = "Latin1_General_CI_AS"
+            queries = [
+                f"""
+            -- Creating theraputics temp table
+            SELECT
+                Patient_ID,
+                CAST(TreatmentStartDate AS DATE) AS TreatmentStartDate,
+                CAST(Received AS DATE) AS Received,
+                Intervention COLLATE {collation} AS Intervention,
+                CurrentStatus COLLATE {collation} AS CurrentStatus,
+                COVID_Indication,
+                Region,
+                MOL1_high_risk_cohort,
+                SOT02_risk_cohorts,
+                CASIM05_risk_cohort
+             INTO {self._therapeutics_table_name} FROM Therapeutics
+            """
+            ]
+        else:
+            queries = []
+        return self._therapeutics_table_name, queries
+
     def patients_with_covid_therapeutics(
         self,
         with_these_statuses=None,
@@ -2801,13 +2827,14 @@ class TPPBackend:
         find_last_match_in_period=None,
         include_date_of_match=False,
     ):
+        therapeutics_table_name, therapeutic_queries = self.create_therapeutics_table()
         filter_conditions = []
         # status may be provided as a single string or a list
         # Available options are: 'Approved','Treatment Complete','Treatment Not Started','Treatment Stopped'
         # Deal with whitespace/case before matching
         status_matches = to_list(with_these_statuses)
         if status_matches:
-            statuses = [quote(status.strip().title()) for status in status_matches]
+            statuses = [quote(status.strip()) for status in status_matches]
             filter_conditions.append(f"CurrentStatus IN ({', '.join(statuses)})")
 
         # Data (Jan 2022) contains the following values:
@@ -2816,7 +2843,7 @@ class TPPBackend:
         therapeutic_matches = to_list(with_these_therapeutics)
         if therapeutic_matches:
             fragments = [
-                f"Intervention COLLATE Latin1_General_CI_AS LIKE {pattern} ESCAPE '!'"
+                f"Intervention LIKE {pattern} ESCAPE '!'"
                 for pattern in codelist_to_like_patterns(
                     therapeutic_matches, prefix="%", suffix="%"
                 )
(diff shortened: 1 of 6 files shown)


cohortextractor/tpp_backend.py
```diff
@@ -2787,6 +2788,187 @@ def patients_outpatient_appointment_date(
...
+            # coalesce the parts (this results in a leading ',')
+            coalesced_parts = " + ".join(
+                f"COALESCE(',' + NULLIF({part}, ''), '')" for part in parts
+            )
+            # string split on ',' to remove duplicates (this results in a second leading ',')
+            split_string = f"""
+            (
+                SELECT DISTINCT ',' + Value FROM STRING_SPLIT({coalesced_parts}, ',')
+                ORDER BY 1 FOR XML PATH('')
+            )
```
rebkwok, line 2910: There was a request to also remove duplicates that may appear in this joined string.  That's fairly simple if TPP was using a modern version of SQL server, but since we can't use `STRING_SPLIT`, it woud involve a complicated XML parse which doesn't seem worth it.  Especially given that removing duplicates should be pretty simple after data extraction, and also currently I don't think there are any patients with risk groups in more than one of these fields, so it's likely to be an infrequent occurrence anyway

cohortextractor/tpp_backend.py
```diff
@@ -2787,6 +2788,180 @@ def patients_outpatient_appointment_date(
...
+                    "SOT02_risk_cohorts",
+                    "CASIM05_risk_cohort",
+                ]
+            ]
+            # coalesce the parts with a leading ','
+            coalesced_parts = " + ".join(
+                f"coalesce(',' + NULLIF({part}, ''), '')" for part in parts
+            )
+            # use stuff to remove the first ','
+            column_definition = f"STUFF({coalesced_parts}, 1, 1, '')"
```
rebkwok, line 2906: There was a request to also remove duplicates that may appear in this joined string. That's fairly simple if TPP was using a modern version of SQL server, but since we can't use STRING_SPLIT, it woud involve a complicated XML parse which doesn't seem worth it. Especially given that removing duplicates should be pretty simple after data extraction, and also currently I don't think there are any patients with risk groups in more than one of these fields, so it's likely to be an infrequent occurrence anyway

cohortextractor/tpp_backend.py
```diff
@@ -2787,6 +2788,207 @@ def patients_outpatient_appointment_date(
...
+    def create_therapeutics_table(self):
+        """
+        Create a temporarary Therapeutics table to use for `with_covid_therapeutics` queries
+        All columns in the Therapeutics table are VARCHAR.  This casts the two date columns that
+        we use in queries to DATE type, and adds case insensitive collation to the Interventaion
+        and CurrentStatus columns.
+        All columns are included, including those that we don't use, so that when we remove
+        duplicates, we only remove complete duplicate rows
+        """
+        if self._therapeutics_table_name is None:
```
evansd, line 2800: Building a temp table for the just this data to get it into the format we want is a neat approach

cohortextractor/tpp_backend.py
```diff
@@ -2787,6 +2788,207 @@ def patients_outpatient_appointment_date(
...
+        # status may be provided as a single string or a list
+        # Available options are: 'Approved','Treatment Complete','Treatment Not Started','Treatment Stopped'
+        # Deal with whitespace/case before matching
+        status_matches = to_list(with_these_statuses)
+        if status_matches:
+            statuses = [quote(status.strip()) for status in status_matches]
+            filter_conditions.append(f"CurrentStatus IN ({', '.join(statuses)})")
+
+        # Data (Jan 2022) contains the following values:
+        # 'Casirivimab and imdevimab '[note trailing space], 'Molnupiravir', 'Remdesivir', 'sarilumab', 'Sotrovimab' , 'Tocilizumab'
```
evansd, line 2855: > [note trailing space]

_Sigh_

cohortextractor/tpp_backend.py
```diff
@@ -2787,6 +2788,207 @@ def patients_outpatient_appointment_date(
...
+        if returning == "binary_flag" or returning == "date":
+            returning = "binary_flag"
+            column_definition = "1"
+        elif returning == "therapeutic":
+            # remove whitespace, convert to comma-separated string
+            column_definition = "REPLACE(LTRIM(RTRIM(Intervention)), ' and ', ',')"
+        elif returning == "risk_group":
+            # First remove any "Patients with a" and replace " and " with "," within individual risk group fields
+            # Then join the 3 risk cohort fields with ","
+            # Note that the last "s" in SOT02_risk_cohorts is correct
```
evansd, line 2906: Also sigh

cohortextractor/tpp_backend.py
```diff
@@ -2787,6 +2788,207 @@ def patients_outpatient_appointment_date(
...
+        else:
+            raise ValueError(f"Unsupported `returning` value: {returning}")
+
+        if use_partition_query:
+            # There can be duplicates per patient in the Therapeutics dataset
+            # These are likely to be invalid in some way, but we keep them in so users
+            # can identify them and deal with them as appropriate
+            # (We remove fully duplicate rows in the temp table only)
+            # Duplicate rows are sorted by all the fields that have been identified to
+            # contain duplicate values for a patient, to ensure a consistent return value
```
evansd, line 2935: I misunderstood this comment at first but I understand now having seen the conversation on the original ticket. So we only ever return one result per patient (as we must) but when we count results we included duplicates and so this is allows researchers to spot that the duplicates are present. Is that right?

evansd (approved):
This all looks good to me. An impressive understanding of all the weird idioms of the cohortextractor! Though also a sad reflection of the state of some of the data we have to work with.

cohortextractor/tpp_backend.py
rebkwok: Yes, so there can be duplicates per patient, which differ on one or more fields, but the assumption (for now at least) is that those duplicates are errors.   Mostly the values are the same apart for one or two fields.   We only ever return one row per patient, either the first or last by TreatmentStartDate, but in the case where there are duplicates, we want the sorting to be consistent.  There's no sensible way to pick which e.g. region is the correct one, if all other fields are the same, so this just makes sure that it always returns the same one (the first, alphabetically).  The idea is that researchers can use the `number_of_matches_in_period` return value to check if they have duplicates and exclude those patients from further analysis.

I'll see if I can make this comment more understandable :)

Issue #614, opened 2021-07-20
ISARIC


sebbacon:
Next steps: TPP to import the data into the OpenSAFELY database.

Currently (morning of 14 October) blocked on conversations / clarifications about linkage, between Chris and the ISARIC team

alexwalkerepi:
There are two tables:

`ISARIC_Patient_Data_TopLine`
 - one row per patient
 - contains a mix of:
   - date columns
   - categorical strings, for example a boolean `"Yes", "No", "(Missing)"` or a unit description such as `"mmol/L", "mg/dL", "(Missing)"`
   - numerical columns

`ISARIC_Patient_Data`
 - multiple rows per patient (~8 per patient on average)
 - contains a similar mix of column types as the topline table, but the column names are not the same.

Both tables are _very_ wide, so I don't think it's practical to write specialised functions to return each specific field. Rather I think that  we should write functions to return the contents of user specified columns. I understand that the data have been pre-processed such that returning the raw values in a column would be fine. There may be a need for date filtering in some cases.

## ISARIC_Patient_Data_TopLine
Being one row per patient already, this feels straightforward, and I would imagine a study definition code would look something like:
### Date variable
```py
    isaric_assessment_date=patients.with_an_isaric_record(
        return_column="symp_or_assess_date",
        return_expectations={"incidence": 0.1, "date": {"earliest": "2020-02-01"}}
    ),
```

### Categorical variable:
```py
    isaric_fever=patients.with_an_isaric_record(
        between=["2020-02-01", "2020-06-01"],
        date_filter_column="dsstdat",
        return_column="fever",
        return_expectations={
            "rate": "universal",
            "category": {
                "ratios": {
                    "Yes": 0.7,
                    "No": 0.1,
                    "(Missing)": 0.2,
                }
            },
        },
    ),
```

### Float variable:
```py
    isaric_sysbp=patients.with_an_isaric_record(
        between=["2020-02-01", "2020-06-01"],
        date_filter_column="dsstdat",
        return_column="sysbp_vsorres",
        return_expectations={
            "incidence": 0.8,
            "float": {"distribution": "normal", "mean": 21, "stddev": 5},
        },
    ),
```

## ISARIC_Patient_Data
This table is multiple rows per patient and many columns are very sparsely populated. I think the pseudocode would likely look very similar to above, however:

[...]

PR #730, opened 2022-02-02
feat: add support for ISARIC

bloodearnest:


cohortextractor/tpp_backend.py
```diff
@@ -2805,6 +2805,35 @@ def patients_with_value_from_file(
...
+        between=None,
+        date_filter_column=None,
+        return_expectations=None,
+    ):
+
+        queries = []
+        table = "ISARIC_Patient_Data_TopLine"
+
+        if between and date_filter_column:
+            date_condition, _ = self.get_date_condition(
```
evansd, line 2820: You need the second argument from this: it's a join condition that tells you what other tables need to be joined to if the date conditions are derived from other queries, rather than being static values.

cohortextractor/tpp_backend.py
```diff
@@ -2805,6 +2805,35 @@ def patients_with_value_from_file(
     def patients_which_exist_in_file(self, f_path):
         return self.patients_with_value_from_file(f_path)
 
+    def patients_with_an_isaric_record(
+        self,
+        returning,
+        between=None,
+        date_filter_column=None,
```
evansd, line 2812: This is unusual but makes sense to me, assuming there are multiple date columns in the ISARIC data that we'd want to filter on. Is there an obvious "primary" date column we could use as a default here? Or is it better to force people to be explicit?

bloodearnest: I'm following @alexwalkerepi suggested api here: https://github.com/opensafely-core/cohort-extractor/issues/614

evansd: :+1: Makes sense

cohortextractor/tpp_backend.py
bloodearnest: fixed

ghickman (approved):

PR #731, opened 2022-02-03
Limit therapeutics risk groups to a verified list

rebkwok:
The Therapeutics dataset comes from a combination of a set of forms completed for a patient, which includes check boxes for the risk group.  One of the forms used to input risk groups contains a free text field.  In the current data, there doesn't appear to be anything that comes from the free text field, but it could appear later, and could potentially mean that disclosive data was accidentally pulled through to the output.  It may (as it should) be being removed before it gets to us, but since we can't be 100% sure of that, we check all extracted risk groups against a verified list of values, and replace anything that isn't in the list with "other".

I've chosen to do this check-and-replace in python after data extraction rather than in SQL when we query the database, because the query for that was getting so complex.

The plan is that we'll query the database when new data is loaded to check for any new values that might need added to the list, which shouldn't happen often, as we're capturing all the ones from the form at the moment.


cohortextractor/therapeutics_utils.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    Helper function to extract distinct risk groups from a file with one line
+    per distinct risk group value extracted from the database
+    """
+    lines = Path(input_file).read_text().split("\n")
+    risk_groups = set()
+    for line in lines:
+        groups = line.split("and")
+        for group in groups:
+            if group:
```
iaindillingham, line 15: Is this because `group == False` when `line == ""`? If so, and the empty lines are at the start or end of the input file, then might this be clearer:

```python
lines = Path(input_file).read_text().strip().split("\n")
```

I'm not yet clear where `input_file` comes from, so sorry for what is probably unnecessary pedantry 🧐.

cohortextractor/tpp_backend.py
```diff
@@ -2788,6 +2824,15 @@ def patients_outpatient_appointment_date(
               OPA.Patient_ID
             """
 
+    @staticmethod
+    def _format_risk_group(original_string):
+        # First remove any "Patients with [a]" and replace " and " with "," within individual risk group fields
```
iaindillingham, line 2829: Are the square brackets significant? (`[a]`)

iaindillingham (approved):
As ever, looking at cohort-extractor makes me hopeful that we'll not have to look at cohort-extractor for much longer. I've read this PR a couple of times and think I understand what it does, how it does it, and why it does it ✅.

I've made some very minor comments; but they're really very minor.

mssql/setup.sql
```diff
@@ -1,4 +1,7 @@
 CREATE DATABASE Test_OpenCorona;
+ALTER DATABASE Test_OpenCorona SET COMPATIBILITY_LEVEL = 100;
```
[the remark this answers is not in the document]
rebkwok: :+1: TPP is actually using SQL Server 2016, but with the compatability level set to 100.  Which is annoying, because we'd get `string_split` with 2016, which would be nice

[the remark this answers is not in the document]
iaindillingham: That is annoying 🙁

cohortextractor/therapeutics_utils.py
rebkwok: Hmm, I'll check.
`input_file` is just a manually generated text file with distinct values from the dataset risk group columns.  Basically, Helen ran a query in the server and copy-pasted the output into a file (after checking that there's nothing there that looks like it came from a free text field and could be disclosive), but it consists of lots of combinations e.g. "liver disease and IMID and solid cancer" etc.  The plan is that someone will run that query when the data changes, and we'll update the `ALLOWED_RISK_GROUPS` with it.  I doubt it'll change often, because the list currently has all the conditions specific on the form checkboxes, but if we do get free text data there could a risk group condition listed there that we want to capture.

cohortextractor/tpp_backend.py
rebkwok: It's just shorthand for "Patients with" and "Patients with a" - on the forms there are checkboxes that list conditions as e.g. "Patients with a haematologic malignancy" and "Patients with renal disease"

iaindillingham: Ah, okay, I see. Thanks for clarifying!

PR #737, opened 2022-02-08
fix: add some better logging of generate_measures command.

bloodearnest:
We need some visibility on memory consumption, and something that might
help up point to troublesome measures in a larger run.

### DIFF
--- a/cohortextractor/cohortextractor.py
+++ b/cohortextractor/cohortextractor.py
@@ -309,8 +309,10 @@ def _generate_measures(
         if date is None:
             continue
         filepath = os.path.join(output_dir, file)
+        logger.info(f"Calculating measures for {filepath}")
         patient_df = None
         for measure in measures:
+            logger.info(f"Calculating {measure.id}")
             output_file = f"{output_dir}/measure_{measure.id}_{date}.csv"
             measure_outputs[measure.id].append(output_file)
             if skip_existing and os.path.exists(output_file):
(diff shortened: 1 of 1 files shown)


cohortextractor/cohortextractor.py
```diff
@@ -319,8 +321,13 @@ def _generate_measures(
             # We do this lazily so that if all corresponding output files
             # already exist we can avoid loading the patient data entirely
             if patient_df is None:
+                logger.info(f"Loading patient data from {filepath}")
                 patient_df = _load_dataframe_for_measures(filepath, measures)
+                logger.info(patient_df.memory_usage())
```
sebbacon, line 326: AFAICT memory_usage is cheap to run, but it may be nicer to do this as `logger.info("Memory %s", patient_df.memory_usage())` so it's not executed at all outside `INFO` level debugging?

ghickman: I don't think the %-formatting would defer it here?  Trying out a function with a `time.sleep(3)` call in the repl appears to execute the function with either type at least, not sure if that's a good enough test though?

bloodearnest: Yeah, we need to pay the cost. It is not free, but compared to minutes to load/calculate the, it's practically free.

Also, the output is a formatted table, rather than just a number, hence why I've printed it wholesale.

bloodearnest: IME, defered string rendering like this is an anti feature, for a few reasons:

1) If you defer execution of things you want, they don't get exercised and can bitrot, and then you turn on debugging logging in prod, and even though logging framework doesn't expose the error, you get garbage exceptions filling up your logs fast. You can't easily catch and handle these errors either.

2) f strings are way more readable

3) If deferring string rendering in logging is a worthwhile performance benefit, then something has gone Very Wrong.

bloodearnest: And in this case, `info` is the default level, so we want it every time anyway.

We are not using logging levels the way ~~java.util.log~~ stdlib was designed, as that is not particularly useful in cli type app like this.

sebbacon: Useful take, thanks!

bloodearnest: This has got missed, can we unblock it?

evansd (approved):

PR #740, opened 2022-02-10
fix: check aggregate date formats recursively

### LINKED ISSUE #739: Not possible include a date in minimum_of that came from another minimum_of
Example:
https://github.com/opensafely/antibody-and-antiviral-deployment/runs/5140032419?check_suite_focus=true

rebkwok:
We check all the date formats in a study definition, including those in an aggregate (e.g. `minimum_of`).  This assumed that all columns in an aggregate would be simple variables, not aggregates themselves.  
Fixes #739


iaindillingham (approved):
Nice!

PR #747, opened 2022-03-07
chore: Update Starburst Trino image to 364-e.4

StevenMaude:
EMIS have updated their Trino version.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,7 @@ services:
     ports:
       - "15785:1433"
   trino:
-    image: "starburstdata/starburst-enterprise:356-e.7"
+    image: "starburstdata/starburst-enterprise:364-e.4"
     ports:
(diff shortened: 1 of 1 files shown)


StevenMaude:
The [version check workflow](https://github.com/opensafely-core/cohort-extractor/actions/runs/1926828020) started failing four days ago. Running the [check workflow on this branch](https://github.com/opensafely-core/cohort-extractor/actions/runs/1944832848) — commit b18edd25a9e9b7ac31462285299f76828cbe29e4 — passes.

evansd (approved):
This change obviously doesn't affect prod at all so happy to merge. But, given the issues @rebkwok uncovered earlier, we should confirm that the upgrade hasn't affected our ability to run queries in prod. (It looks like the last extraction we ran was `2022-02-28 18:19:39+0000`, so before this change.)

PR #748, opened 2022-03-07
Support returning total bed days and total critical care days from patients.admitted_to_hospital

### LINKED ISSUE #742: APCS: support returning "total bed days"
### Use case
**Total bed days** can be used as a measure of healthcare usage, or total amount of time spent in hospital over a particular period with a certain diagnosis, even if it was across multiple spells.

### Current workaround
It is difficult to calculate total bed days per patient over a given period of time where they may have had multiple spells. Currently it would be required to extract a long sequence of admission and discharge dates over the entire period, calculate the difference between each pair and then sum, but may be imperfect, e.g. if someone has more admissions in the period than we have allowed for. 

### Existing options in this dataset which are implemented already
We currently offer the following somewhat similar `returning` options: 
- `number_of_matches_in_period` - number of times patient was admitted in time period specified.
- `days_in_critical_care` - number of days in critical care during spell.

### Suggested implementation
Should be much the same as `days_in_critical_care`:
Sum of `Der_Spell_LoS` [length of stay] for all included spells in period (we also have another field available, `Spell_LoS` in one of the linked APCS tables but I've confirmed they are identical). Required as a `returning` option only.
- However, `LoS` is measured in number of _nights_ not number of days so single-day admissions return 0, a two-day stay returns 1, and so on. Could have two options: `length_of_stay_nights` (using the field as it is) or `length_of_stay_days` (adding 1 to each value)

### Notes/issues
- If someone has a single-day admission and a longer stay both starting the same day, we would count the first day twice unless we account for this somehow. However this is relatively unusual.
- There are also a very small number of duplicates / other impossible overlapping spells so this would very occasionally return spurious results.

rebkwok:
Fixes #742 

`patients.admitted_to_hospital` already returns `days_in_critical_care`, however this will return the number of days for the first or last matching admission.  

This allows the return options of `total_critical_care_days_in_period` and `total_bed_days_in_period` by summing the days for all admissions during the period.  In the case where there are two or more admissions recorded with the same admission date, we take the max number of days to avoid counting the sames days more than once.

For `total_bed_days_in_period`, the return value required is number of days, but the data recorded is number of nights (sometimes 0, which counts as one bed day).

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -2387,6 +2387,7 @@ def admitted_to_hospital(
             * `days_in_critical_care`: number of days in critical care during spell,
             * `administrative_category`: private vs NHS funded treatment,
             * `duration_of_elective_wait`: days on waiting list for planned procedures (use with caution).
+            * `total_bed_days_in_period`: total number of bed days for all admissisions during the time period specified
 
         find_first_match_in_period: a boolean that indicates if the data returned is first event
             if there are multiple matches within the time period
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -2387,7 +2387,8 @@ def admitted_to_hospital(
             * `days_in_critical_care`: number of days in critical care during spell,
             * `administrative_category`: private vs NHS funded treatment,
             * `duration_of_elective_wait`: days on waiting list for planned procedures (use with caution).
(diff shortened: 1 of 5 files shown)


iaindillingham (approved):
This is really clear. Thanks for commenting on the adjustment, and making it explicit in the code.

PR #749, opened 2022-03-10
fix: disambiguate column references in emis registered_as_of

rebkwok:
This allows a date variable to be passed to `registered_as_of` in the emis backend, e.g. to get all patients registered on the date of death as recorded in ONS (with the caveat that EMIS registrations are limited to just the current/latest registration):

```
study = StudyDefinition(
        ons_death_date=patients.died_from_any_cause(
            returning="date_of_death", date_format="YYYY-MM-DD"
        ),
        population=patients.registered_as_of("ons_death_date"),
)
```


inglesp (approved):
Aaargh, well done.

PR #753, opened 2022-03-14
dependency cleanup

bloodearnest:
- chore: remove requests_pkcs dependencies and its children.
- chore: move sqlachemy to being a dev only dependency
- chore: remove uneeded dependencies

### DIFF
--- a/requirements.prod.txt
+++ b/requirements.prod.txt
@@ -71,7 +71,5 @@ structlog==20.2.0
     # via opensafely-cohort-extractor (setup.py)
 tabulate==0.8.7
     # via opensafely-cohort-extractor (setup.py)
-tinynetrc==1.3.0
-    # via opensafely-cohort-extractor (setup.py)
 urllib3==1.26.5
     # via requests
--- a/requirements.prod.txt
+++ b/requirements.prod.txt
@@ -65,8 +65,6 @@ six==1.14.0
     # via
     #   presto-python-client
     #   python-dateutil
-sqlalchemy==1.3.15
-    # via opensafely-cohort-extractor (setup.py)
 sqlparse==0.3.1
     # via opensafely-cohort-extractor (setup.py)
 structlog==20.2.0
--- a/requirements.prod.txt
+++ b/requirements.prod.txt
@@ -6,14 +6,10 @@
 #
 certifi==2020.4.5.1
     # via requests
-cffi==1.14.4
-    # via cryptography
 chardet==3.0.4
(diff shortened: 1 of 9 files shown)


requirements.prod.txt
```diff
@@ -41,14 +35,10 @@ presto-python-client==0.7.0
     # via opensafely-cohort-extractor (setup.py)
 py==1.10.0
     # via retry
-pyarrow==3.0.0
+pyarrow==7.0.0
```
ghickman, line 38: Did you mean to bump this?

ghickman (approved):
Always love some dependencies being removed 👌

requirements.prod.txt
bloodearnest: No, was accidental, I tried removing it but it was needed by some code
 will revert

setup.py
```diff
@@ -17,17 +17,15 @@
     author_email="tech@opensafely.org",
     python_requires=">=3.7",
     install_requires=[
-        "lz4",
```
evansd, line 20: I seem to remember this is needed for compressed feather files. Once we confirm the situation we should add a comment here to explain.

evansd (changes requested):
So pleased about this! We'll need to confirm the status of `lz4` first though.

setup.py
bloodearnest: Ah, ok, let me dig into it.

There were no tests failing when removing it, so...

bloodearnest: https://arrow.apache.org/docs/python/generated/pyarrow.feather.write_feather.html#pyarrow.feather.write_feather

Note that support for that was added in pandas 1.1.0, we have 1.0.1, so we had the dependency, but it wasn't actually being used!

I will restore it, I'm doing an upgrade PR next.

evansd: I _think_ that, helpfully, it might just write uncompressed files if LZ4 isn't available:
https://arrow.apache.org/docs/python/generated/pyarrow.feather.write_feather.html#pyarrow.feather.write_feather

bloodearnest: Ah, no https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_feather.html the kwargs was unsupported, but the underlying defaults were valid.

Anyway, will restore, and am writing functional tests for these formats to make sure we can write all the formats as expected.

bloodearnest: restored

evansd (approved):

PR #760, opened 2022-03-16
fix: fix dta exports and feather compression

bloodearnest:
* bugfix .dta.gz export's were not actually gzip compressed. Upgraded
  pandas to enable this.

* used the more recent version 118 dta format, which supports unicode
  and longer strings.

* feather compression switched to using zstd as it is both smaller on
  disk and equivalent speed to decompress.  Also doesn't require a the
  lz4 dependency, so was removed.

* wrote tests to excercise the above exports

### DIFF
--- a/cohortextractor/pandas_utils.py
+++ b/cohortextractor/pandas_utils.py
@@ -13,7 +13,8 @@ def dataframe_to_file(df, filename):
     if filename.endswith(".csv") or filename.endswith(".csv.gz"):
         df.to_csv(filename, index=False)
     elif filename.endswith(".feather"):
-        df.to_feather(filename)
+        # zstd is a better option that the default lz4 compression
+        df.to_feather(filename, compression="zstd")
     elif filename.endswith(".dta") or filename.endswith(".dta.gz"):
         # Make a shallow copy of the dataframe so we can replace certain
         # columns without duplicating the entire thing in memory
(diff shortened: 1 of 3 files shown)


cohortextractor/pandas_utils.py
```diff
@@ -13,7 +13,8 @@ def dataframe_to_file(df, filename):
     if filename.endswith(".csv") or filename.endswith(".csv.gz"):
         df.to_csv(filename, index=False)
     elif filename.endswith(".feather"):
-        df.to_feather(filename)
+        # zstd is a better option that the default lz4 compression
```
evansd, line 16: Sounds sensible. Have we confirmed that our default R image can read these?

setup.py
```diff
@@ -17,7 +17,6 @@
     author_email="tech@opensafely.org",
     python_requires=">=3.7",
     install_requires=[
-        "lz4",  # implicit import for compressed feather files
```
evansd, line 20: I worry this means that people won't be able to create measures from previous feather extracts. I think sadly we might need to leave it in for a bit until we've confirmed that no one needs to do this.

Dockerfile
```diff
@@ -122,6 +122,8 @@ RUN cohortextractor --help
 # Development image that includes test dependencies and mounts the code in
 FROM cohortextractor-base as cohortextractor-dev
 
+RUN --mount=type=cache,target=/var/cache/apt /root/docker-apt-install.sh libmagic1
```
evansd, line 125: How much is libmagic actually getting us here, given that it doesn't understand either feather or dta? It seems like we're basically just using it to confirm that gzipped data is actually gzipped, which we could do with stdlib code if we wanted to, right?

evansd (approved):
Looks good. Approved modulo that question about LZ4 support.

cohortextractor/pandas_utils.py
bloodearnest: Yes

setup.py
bloodearnest: Curse you lz4. This day you survive again, but I will kill you one day.

Dockerfile
bloodearnest: fair point, it makes the test a bit awkward, but probably worth it

PR #762, opened 2022-03-22
feat: Support new SGTF field in SGSS data

evansd:
Note: this is based on an assumption about what the data will look like when we actually get it, but we'll need to wait for that to come through before we can verify this and merge the PR (hence marking it as draft).


inglesp (approved):
Took me a little while to work out what was going on here!

PR #764, opened 2022-03-24
Support fixed dates in `minimum_of/maximum_of`

### LINKED ISSUE #695: Using fixed and dynamic dates with patients.maximum_of()
Our study defines the start of follow-up for each individual as the latest of study start date (a fixed date) and 2 weeks after second vaccination (a dynamic date that differs for each individual). We therefore wanted to derive a date variable that reflected this and could then be used to define the exposures, outcomes and covariates for the study. 

I tried using `patients.maximum_of()` but it doesn't like the combination of a fixed and dynamic dates.  Assuming this is an issue and I haven't missed something, I think potential solutions are to add functionality that would define a variable equal to index_date for everyone that could then be fed to `patients.maximum_of()` or to update `patients.maximum_of()` to allow combinations of a fixed and dynamic dates.

evansd:
Closes #695


iaindillingham (approved):
As Dirty Harry say, "A man's got to know his limitations." And with cohort-extractor, I think I know mine. I think I can see what's happening, and why it's happening, for dates (i.e. strings that conform to the ISO date format). What would happen for other types? If I added an aggregate-of query, but passed an integer, I sense cohort-extractor would "just work". I can't see how, though.

iaindillingham:
Getting there 🙂  In principle, then, we _could_ add syntactic sugar for non-dates; in practice, we have chosen not to, because it's not needed (#695) and because there might be confusion between non-dates and column names, making it harder to implement. If we did add it, then would we add an `elif` here:

https://github.com/opensafely-core/cohort-extractor/blob/f29af617747e862969e1e3105246a993e7ddebdd/cohortextractor/process_covariate_definitions.py#L460-L473

PR #765, opened 2022-03-25
update flake8 to match our template

ghickman:
As per opensafely-core/repo-template#51


StevenMaude:
question: Why does the repo-template have this [additional line](https://github.com/opensafely-core/repo-template/blob/af24a14848df188d258680abb412ae287ec894dc/.flake8#L8), but not here?

StevenMaude:
This looks fine besides :+1:

StevenMaude (approved):

PR #767, opened 2022-03-30
chore: apply black fixes

ghickman:
Not sure why these weren't picked up locally or in CI for #766 so here we are


lucyb (approved):

evansd:
I think this might have been part of the Github Actions "special time". It looks like only the Semantic PR check ran on the previous PR and nothing else:
github com_opensafely-core_cohort-extractor_pull_766

ghickman:
_sigh_

Yep, that will be it for CI!  I'm still surprised pre-commit didn't pick it up locally though 🤔

evansd:
I think that makes sense: pre-commit runs before the commit, so a commit which upgrades the version of Black is going to get checked with the previous version of Black.

ghickman:
#766 includes a change to the pre-commit config though, that would have run with the updated version

PR #773, opened 2022-04-05
Add support for ONS_CIS

### LINKED ISSUE #587: Developing backend functionality to analysis ONS CIS extract
ONS is now sharing ONS Community Infection Survey (ONS-CIS) data with TPP - see [here](https://github.com/opensafely/database-notebooks/blob/master/notebooks/database-builds.ipynb).

ONS are having regular governance calls with OpenSAFELY to decide which studies to conduct against this ONS-CIS dataset.

Importantly:
1. ONS is the decision maker on which studies can be analysed against this. 
2. The ONS-CIS data must ONLY be processed if the ONS governance board approve the project

We are due to identify a small handful of projects soon. 

@alexwalkercebm is on the discussion committee alongside Laurie Tomlinson.
@sebbacon is there a particular development workflow we need to use to go from ingestion; backend; analyse?

rebkwok:
Fixes #587 

Adds `patients_with_an_ons_cis_record` method to the TPP backend.

Similar to ISARIC, this method allows returning any column from the dataset.  There are multiple rows per patient (more or less unique by visit_id and visit_date), but as there are multiple date columns, we also have a `date_filter_column` that can be specified.

Many of the columns are coded - e.g. sex takes values of 1 or 2, where 1=male and 2=female.  The keys are numeric, and mostly these columns are integer type, although a few are varchar. (but still numbers, stringified).  Unless explicitly requested, we turn these codes into their string labels in the returned data.  If the original codes are requested (with `return_category_labels=False`), we have to cast them to varchar, otherwise any integer-type values will return 0 when there's no data, which is often a valid code itself.

See also corresponding PRS:
restrict use of ONS_CIS dataset: https://github.com/opensafely-core/opensafely-cli/pull/111
docs: https://github.com/opensafely/documentation/pull/677


rebkwok:
Note [this comment](https://github.com/opensafely-core/cohort-extractor/issues/587#issuecomment-1088733829)

There are rows that have the same visit ID for more than one patient.  Where that happens, it tends to happen several times for same 2 patients across multiple different visits, which seems like it is valid, not just a data entry error.  Possibly a form can be completed once for a household, not just for a single patient - waiting for confirmation from TPP

tests/test_tpp_backend.py
```diff
@@ -5858,3 +5859,152 @@ def get_cursor(self):
...
+            date_filter_column="covid_test_blood_pos_first_date",
+            on_or_before="2021-07-31",
+        ),
+        # boolean value
+        self_isolating_at_last_visit=patients.with_an_ons_cis_record(
+            returning="self_isolating"
+        ),
+        # Coded categories are converted to long-form string labels
+        # result_tdi is a coded category value; varchar type in the db
+        # raw values are stringified ints, return the long form category lable
```
iaindillingham, line 5968:
```suggestion
        # raw values are stringified ints, return the long form category label
```

tests/test_tpp_backend.py
```diff
@@ -5858,3 +5859,152 @@ def get_cursor(self):
...
+        result_tdi=patients.with_an_ons_cis_record(
+            returning="result_tdi",
+            find_first_match_in_period=True,
+            include_date_of_match=True,
+            date_format="YYYY-MM-DD",
+        ),
+        # country is a coded category value, int type in the db
+        # raw values are ints, return the long form category lable
```
iaindillingham, line 5976:
```suggestion
        # raw values are ints, return the long form category label
```

tests/test_tpp_backend.py
```diff
@@ -5858,3 +5859,152 @@ def get_cursor(self):
...
+
+
+def test_ons_cis():
+    session = make_session()
+    session.add_all(
+        [
+            Patient(
+                ONS_CIS=[
+                    ONS_CIS(
+                        age_at_visit="20",
```
iaindillingham, line 5871: Is this a string for testing some aspect of the method? We use `int`s elsewhere.

```suggestion
                        age_at_visit=20,
```

tests/test_tpp_backend.py
```diff
@@ -5858,3 +5859,152 @@ def get_cursor(self):
...
+                    ),
+                    ONS_CIS(
+                        age_at_visit=41,
+                        visit_date="2021-10-01",
+                        covid_test_blood_pos_first_date="2021-06-01",
+                        result_tdi="1",
+                        country=2,
+                        self_isolating=0,
+                    ),
+                    # duplicate record ignored in number_of_matches_in_period counts
```
iaindillingham, line 5910: Thanks for commenting here. It saved me a lot of head-scratching 🙂

tests/test_tpp_backend.py
```diff
@@ -5858,3 +5859,152 @@ def get_cursor(self):
...
+            find_first_match_in_period=True,
+            date_filter_column="covid_test_blood_pos_first_date",
+            on_or_after="2021-06-15",
+        ),
+    )
+
+    assert_results(
+        study.to_dicts(convert_to_strings=False),
+        age_at_visit=[20, 30, 41],
+        age_at_visit_date=["2021-10-01", "2021-10-01", "2021-10-01"],
```
iaindillingham, line 5997: Thanks for returning the dates here (and elsewhere). It made associating the fixtures with the results much easier. It's also a helpful reminder that `visit_date` is the default date column.

iaindillingham (approved):
A couple of typo suggestions, but that's all. Really clear to read and reason about 🙂

tests/test_tpp_backend.py
rebkwok: No, it's an int in the raw table, will update it

rebkwok:
Thanks @iaindillingham - I'm having some second thoughts on whether it's sensible to determine the `date_filter_column` based on the `returning` value, and whether we should either always use the same default, or require users to explicitly specifiy which `date_filter_column` they want; added a comment to the issue, so I'll wait for feedback before merging this one

iaindillingham (approved):

Issue #494, opened 2021-03-15
`codelist` functiion should ensure it hasn't been passed a string


evansd:
If we accidentally write `codelist(code, system="snomed")` rather than `codelist([code], system="snomed")` then `code` will be implicitly converted to a list of characters, and all the tests will pass, but the behaviour will be not at all what's intended in production (see https://github.com/opensafely-core/cohort-extractor/issues/493#issuecomment-799501197).

We should prevent this, either by explicitly throwing an error if given or a string or when given anything that isn't a list, tuple or Codelist instance.

PR #780, opened 2022-04-11
fix: convert a single string code to a list

rebkwok:
A user wants to match a single code, and provides the argument as a string rather than a list, e.g.
```
rheum_appt_date=patients.outpatient_appointment_date(
        returning="date",
        with_these_treatment_function_codes='410',
        date_format="YYYY-MM-DD",
        between=["index_date - 1 year", "index_date"],
        return_expectations={
            "incidence": 0.9,
            "date": {"earliest": year_preceding, "latest": end_date},
        },
    )
```

Previously we assumed (but didn't enforce) that the `with_these_treatment_function_codes` argument is either a codelist, or a list of codes that we can convert into a codelist.  Calling `list(codelist)` on a string argument doesn't raise an error, instead it converted the single code into a list of characters - i.e. `["4", "1", "0"]` in the above example, and opaquely failed to match any 
records in the real data.

Since providing a single code as a string seems like a reasonable thing for a user to expect to be able to do, we now just check if it's a string before converting to a list.

### DIFF
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -3524,6 +3524,8 @@ def codelist_to_list(codelist):
     if getattr(codelist, "has_categories", False):
         return [code for (code, category) in codelist]
     else:
+        if isinstance(codelist, str):
+            return [codelist]
         return list(codelist)
 
 
--- a/tests/test_tpp_backend.py
+++ b/tests/test_tpp_backend.py
@@ -4585,8 +4585,14 @@ def test_outpatient_appointment_date_returning_binary_flag_with_these_treatment_
         opa=patients.outpatient_appointment_date(
             returning="binary_flag", with_these_treatment_function_codes=["812", "813"]
         ),
+        # we can match a single code, provided as a string
+        opa1=patients.outpatient_appointment_date(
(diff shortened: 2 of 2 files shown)


rebkwok:
I'm somewhat torn between this fix, and checking the type of the codelist argument and raising an exception if it's not a codelist or an iterable; but doing `with_these_codes="123"` seems like a logical thing for a user to do with a single code.

evansd (approved):
I think this is a reasonable solution, thanks for picking it up.

There's an issue which has been open for over a year to do the other, stricter option you mentioned but we've never done it:
https://github.com/opensafely-core/cohort-extractor/issues/494

I think I would previously have advocated for being stricter (and really I suppose we should be enforcing that the code comes from the right coding system) but I think that ship's long sailed with cohortextractor and hopefully we can produce something both stricter and more usable with databuilder.

rebkwok:
> I think I would previously have advocated for being stricter (and really I suppose we should be enforcing that the code comes from the right coding system) but I think that ship's long sailed with cohortextractor and hopefully we can produce something both stricter and more usable with databuilder.

Yes, that was more or less my thoughts too. Where possible, I'm trying to do the simplest thing when we need to add things to cohort-extractor now.

PR #781, opened 2022-04-11
Renal reg

CarolineMorton:
Start of adding the renal registry to the backend.

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -2858,3 +2858,73 @@ def with_an_ons_cis_record(
             )
     """
     return "with_an_ons_cis_record", locals()
+
+
+
+def with_record_in_UKRR(
+    # picks dataset held by UKRR
+    from_dataset,
+    returning,
+    # Date filtering: date limits
+    on_or_before=None,
+    on_or_after=None,
+    between=None,
+    # Matching rule
+    find_first_match_in_period=None,
+    find_last_match_in_period=None,
+    include_date_of_match=False,
+    date_format=None,
+    return_expectations=None,
+):
+    """
+    Return whether patient has a record in the UK Renal Registry
+
+    Args:
+        from_dataset: string value; options are:
+            * '2019_prevalence' - a prevalence cohort of patients alive and on RRT in December 2019
+            * '2020_prevalence' - a prevalence cohort of patients alive and on RRT in December 2020
+            * '2021_prevalence' - a prevalence cohort of patients alive and on RRT in December 2021
+            * '2020_incidence' - an incidence cohort of patients who started RRT in 2020
+            * '2020_ckd' - a snapshot prevalence cohort of patient with Stage 4 or 5 CKD who were
+                reported to the UKRR to be under renal care in December 2020.
+        returning: string value; options are:
+            * "binary_flag"
+            * "number_of_matches_in_period"
+            * "renal_centre" - string indicating the code of the main renal centre a
+                patient is registered with
+            * "rrt_start_date" - the latest start date for renal replacement therapy
+            * "treatment_modality_start" - the treatment modality at `rrt_start_date` such as
+                ICHD, HHD, HD, PD, Tx
+            * "treatment_modality_prevalence" - the treatment modality from the prevalence data
+            * "latest_creatinine" - most recent creatinine held by UKRR
+            * "latest_egfr" - most recent eGFR held by UKRR
+        on_or_before: date of interest as a string with the format `YYYY-MM-DD`. Filters results to measurements
+            on or before the given date (as defined by `date_filter_column`).
+        on_or_after: date of interest as a string with the format `YYYY-MM-DD`. Filters results to measurements
+            on or after the given date (as defined by `date_filter_column`).
+        between: two dates of interest as a list with each date as a string with the format `YYYY-MM-DD`.
+            Filters results to measurements between the two dates (as defined by `date_filter_column`)
+            provided (inclusive).
+            The two dates must be in chronological order.
+        find_first_match_in_period: as described elsewhere
+        find_last_match_in_period: as described elsewhere
+        include_date_of_match: a boolean indicating if an extra column containing the date (from `date_filter_column`)
+            of the match should be returned.
+        date_format: a string detailing the format of dates to be returned.
+            It can be "YYYY-MM-DD", "YYYY-MM" or "YYYY" and wherever possible the least disclosive data should be
+            returned. i.e returning only year is less disclosive than a date with month and year.
+        return_expectations: as described elsewhere.
+
+    Example:
+        Return patients who are in the prevalence dataset of the UKRR in 2019.
+
+            ukrr_2019 = patients.with_record_in_UKRR(
+                from_dataset="2019_prevalence',
+                returning="binary_flag",
+                find_first_match_in_period=True,
+                return_expectations={
+                    "incidence": 0.25
+                },
+            )
+    """
+    return "with_record_in_UKRR", locals()
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -3492,6 +3492,24 @@ class TPPBackend:
         return table_queries + [sql]
 
 
+def patients_with_record_in_UKRR(
+        # picks dataset held by UKRR
+        from_dataset,
+        returning,
+        # Date filtering: date limits
+        on_or_before=None,
+        on_or_after=None,
+        between=None,
+        # Matching rule
+        find_first_match_in_period=None,
+        find_last_match_in_period=None,
+        include_date_of_match=False,
+        date_format=None,
+        return_expectations=None,
+):
+    pass
+
+
 class ColumnExpression:
     def __init__(
         self,
--- a/tests/test_tpp_backend.py
+++ b/tests/test_tpp_backend.py
@@ -59,6 +59,7 @@ from tests.tpp_backend_setup import (
     SGSS_Negative,
     SGSS_Positive,
     Therapeutics,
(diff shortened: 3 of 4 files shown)


cohortextractor/patients.py
```diff
@@ -2858,3 +2858,73 @@ def with_an_ons_cis_record(
             )
     """
     return "with_an_ons_cis_record", locals()
+
+
+
+def with_record_in_UKRR(
+    # picks dataset held by UKRR
+    from_dataset,
+    returning,
```
rebkwok, line 2867: pretty much everywhere else in patients.py we use keyword args only - I'm assuming that's to prevent users putting positional args in the wrong order

rebkwok (commented):
Looks good so far - just a few minor comments

cohortextractor/patients.py
```diff
@@ -2858,3 +2858,73 @@ def with_an_ons_cis_record(
...
+            * "binary_flag"
+            * "number_of_matches_in_period"
+            * "renal_centre" - string indicating the code of the main renal centre a
+                patient is registered with
+            * "rrt_start_date" - the latest start date for renal replacement therapy
+            * "treatment_modality_start" - the treatment modality at `rrt_start_date` such as
+                ICHD, HHD, HD, PD, Tx
+            * "treatment_modality_prevalence" - the treatment modality from the prevalence data
+            * "latest_creatinine" - most recent creatinine held by UKRR
+            * "latest_egfr" - most recent eGFR held by UKRR
```
rebkwok, line 2900: If it makes more sense, we can map these things to more user-friendly labels (same for the dataset names).  I'm not sure if it's helpful here, but some of the other methods do (e.g. `patients_admitted_to_hospital`).

cohortextractor/patients.py
```diff
@@ -2858,3 +2858,73 @@ def with_an_ons_cis_record(
...
+            on or before the given date (as defined by `date_filter_column`).
+        on_or_after: date of interest as a string with the format `YYYY-MM-DD`. Filters results to measurements
+            on or after the given date (as defined by `date_filter_column`).
+        between: two dates of interest as a list with each date as a string with the format `YYYY-MM-DD`.
+            Filters results to measurements between the two dates (as defined by `date_filter_column`)
+            provided (inclusive).
+            The two dates must be in chronological order.
+        find_first_match_in_period: as described elsewhere
+        find_last_match_in_period: as described elsewhere
+        include_date_of_match: a boolean indicating if an extra column containing the date (from `date_filter_column`)
```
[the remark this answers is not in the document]
CarolineMorton: agree

cohortextractor/patients.py
CarolineMorton: I'm not sure I completely understand this comment. Do you mean change the order of them or remove? We need to be able to identify the dataset in the UKRR somehow.

tests/tpp_backend_setup.py
```diff
@@ -1012,3 +1012,21 @@ class ONS_CIS(Base):
...
+
+
+class UKRR(Base):
+    __tablename__ = "UKRR"
+
+    # fake pk to satisfy the ORM
+    id = Column(Integer, primary_key=True)
+
+    Patient_ID = Column(Integer, ForeignKey("Patient.Patient_ID"))
+    Patient = relationship("Patient", back_populates="UK_RR")
```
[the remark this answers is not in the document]
CarolineMorton: aye - typo!

cohortextractor/patients.py
```diff
@@ -2858,3 +2858,73 @@ def with_an_ons_cis_record(
...
+    # picks dataset held by UKRR
+    from_dataset,
+    returning,
+    # Date filtering: date limits
+    on_or_before=None,
+    on_or_after=None,
+    between=None,
+    # Matching rule
+    find_first_match_in_period=None,
+    find_last_match_in_period=None,
```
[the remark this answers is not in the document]
CarolineMorton: thanks

cohortextractor/patients.py
rebkwok: I just meant to make them keywords instead of positional args.  It means users can't mix up the order and specify `returning` where we think we're getting `from_dataset`.  

Like:
```
def with_record_in_UKRR(
    # picks dataset held by UKRR
    from_dataset=None,
    returning="binary_flag",
    ...
```

I think elsewhere we generally default to `binary_flag` for `returning`.  `from_dataset` could have a default if it makes sense to default to one of them, or None and we raise an exception if they don't provide it

cohortextractor/patients.py
```diff
@@ -2858,3 +2858,62 @@ def with_an_ons_cis_record(
...
+            * "renal_centre" - string indicating the code of the main renal centre a
+                patient is registered with
+            * "rrt_start_date" - the latest start date for renal replacement therapy
+            * "treatment_modality_start" - the treatment modality at `rrt_start_date` such as
+                ICHD, HHD, HD, PD, Tx
+            * "treatment_modality_prevalence" - the treatment modality from the prevalence data
+            * "latest_creatinine" - most recent creatinine held by UKRR
+            * "latest_egfr" - most recent eGFR held by UKRR
+        on_or_before: date of interest as a string with the format `YYYY-MM-DD`. Filters results to measurements
+            on or before the given date (as defined by `date_filter_column`).
```
rebkwok, line 2896: copy/paste error (and in the next 2 args)?  There's no date_filter_column

cohortextractor/tpp_backend.py
```diff
@@ -3491,6 +3491,45 @@ def patients_with_an_ons_cis_record(
...
+    def patients_with_record_in_ukrr(
+        self,
+        # picks dataset held by UKRR
+        from_dataset,
+        returning,
+        # Date filtering: date limits
+        on_or_before=None,
+        on_or_after=None,
+        between=None,
+        include_date_of_match=False,
```
rebkwok, line 3503: remove this arg?

cohortextractor/tpp_backend.py
```diff
@@ -3491,6 +3491,45 @@ def patients_with_an_ons_cis_record(
...
 
+    def patients_with_record_in_ukrr(
+        self,
+        # picks dataset held by UKRR
+        from_dataset,
+        returning,
+        # Date filtering: date limits
+        on_or_before=None,
+        on_or_after=None,
+        between=None,
```
rebkwok, line 3502: We only need `between` here - `process_covariate_definitions.handle_time_period_options` converts any `on_or_before` or `on_or_after` args into a between one before it gets to the backend.

cohortextractor/tpp_backend.py
```diff
@@ -3491,6 +3491,45 @@ def patients_with_an_ons_cis_record(
...
+        elif returning == "treatment_modality_start":
+            column_definition = "mod_start"
+        elif returning == "treatment_modality_prevalence":
+            column_definition = "mod_prev"
+        elif returning == "latest_creatinine":
+            column_definition = "creat"
+        elif returning == "latest_egfr":
+            column_definition = "eGFR_ckdepi"
+        else:
+            column_definition = returning
```
rebkwok, line 3525: This could be nicer as another mapping:

```
column_mapping = {
    "binary_flag": 1,
    "treatment_modality_start": "mod_start",
    ...
}
column_definition = column_mapping.get(returning)
```

And then we can also check and raise an exception if the `returning` value isn't one we expected

tests/test_tpp_backend.py
```diff
@@ -6021,3 +6025,100 @@ def test_ons_cis():
...
+    session.add_all(
+        [
+            Patient(
+                UKRR=[
+                    UKRR(
+                        dataset="2019prev",
+                        renal_centre="AD100",
+                        rrt_start="2019-01-01",
+                        mod_start="RT",
+                        mod_prev="NULL",
```
rebkwok, line 6042: I'm not entirely confident about this, but I think you want None here, and sqlalchemy will make it a null in the db.  (Unless nulls in the real data are represented by the string "NULL".)

rebkwok (changes requested):
I left a few minor suggestions/questions, but this looks good :+1: 

THere's something formatting-related that black is complaining about atm

tests/test_tpp_backend.py
```diff
@@ -6021,3 +6025,198 @@ def test_ons_cis():
...
+                        eGFR_ckdepi=1.588,
+                    ),
+                ],
+            ),
+            Patient(
+                UKRR=[
+                    UKRR(
+                        dataset="2019prev",
+                        renal_centre="AD100",
+                        rrt_start="2019-10-01",
```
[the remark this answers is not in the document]
CarolineMorton: Yes this is possible. I have double checked. Some patients just went straight to transplant.

tests/test_tpp_backend.py
```diff
@@ -6021,3 +6025,198 @@ def test_ons_cis():
...
         country_as_codes=["0", "1", ""],
     )
+
+
+def test_ukrr():
+    "Test UK Renal Registry"
+    session = make_session()
+    session.add_all(
+        [
+            Patient(
```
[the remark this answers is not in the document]
CarolineMorton: thanks I have added these comments more to the test study definitions as it makes more sense there

tests/test_tpp_backend.py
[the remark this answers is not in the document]
rebkwok: We should probably mention this somewhere - maybe just in the dataset description in the docs though

rebkwok (approved):

tests/test_tpp_backend.py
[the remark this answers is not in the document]
CarolineMorton: I will put this in the documentation I think,

PR #785, opened 2022-04-21
feat: Add generate_codelist_report

inglesp:
generate_codelist_report generates a pair of CSV files reporting on
usage of codes in a given codelist.

Paired on with @lucyb.

### DIFF
--- a/cohortextractor/generate_codelist_report.py
+++ b/cohortextractor/generate_codelist_report.py
@@ -44,8 +44,8 @@ def generate_codelist_report(output_dir, codelist_path, start_date, end_date):
 
     os.makedirs(output_dir, exist_ok=True)
 
-    counts_per_code.to_csv(output_dir / "counts_per_code.csv")
-    counts_per_week.to_csv(output_dir / "counts_per_week.csv")
+    counts_per_code.to_csv(os.path.join(output_dir, "counts_per_code.csv"))
+    counts_per_week.to_csv(os.path.join(output_dir, "counts_per_week.csv"))
 
 
 def codelist_queries(codelist):
--- a/cohortextractor/generate_codelist_report.py
+++ b/cohortextractor/generate_codelist_report.py
@@ -1,0 +1,111 @@
+import os
+
+import pandas as pd
+import structlog
+
+from .codelistlib import codelist_from_csv
+from .mssql_utils import mssql_dbapi_connection_from_url
+from .tpp_backend import make_batches_of_insert_statements, quote
+
+logger = structlog.get_logger()
+
+
+def generate_codelist_report(output_dir, codelist_path, start_date, end_date):
+    """
+    Generate a pair of CSV files reporting usage of codes in given codelist.
+
+        * counts_per_code counts how many times each code appears in the given timeframe
+        * counts_per_week counts how many times all the codes appear in each week in the
+            given timeframe
+
+    Weeks begin on a Monday, and a code appearing on (say) a Wednesday is counted as
+    belonging to the following Monday.
+    """
+
+    codelist = codelist_from_csv(codelist_path, "snomedct")
+
+    queries = codelist_queries(codelist) + [
+        events_query(start_date, end_date),
+        population_query(start_date, end_date),
+        results_query(),
+    ]
+
+    connection = mssql_dbapi_connection_from_url(os.environ["DATABASE_URL"])
+    cursor = connection.cursor()
+
+    for query in queries:
+        logger.debug(query)
+        cursor.execute(query)
+
+    counts = pd.DataFrame(list(cursor), columns=["code", "date", "num"])
+
+    counts_per_code = counts.groupby("code")["num"].sum()
+    counts_per_week = counts.groupby(pd.Grouper(key="date", freq="W-MON"))["num"].sum()
+
+    os.makedirs(output_dir, exist_ok=True)
+
+    counts_per_code.to_csv(output_dir / "counts_per_code.csv")
+    counts_per_week.to_csv(output_dir / "counts_per_week.csv")
+
+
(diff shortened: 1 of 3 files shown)


cohortextractor/generate_codelist_report.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def events_query(start_date, end_date):
+    return f"""
+    SELECT * INTO #events FROM (
+        SELECT Patient_ID, ConceptID, ConsultationDate
+        FROM CodedEvent_SNOMED
+        INNER JOIN #codelist
+            ON CodedEvent_SNOMED.ConceptID = #codelist.code
+        WHERE ConsultationDate BETWEEN '{start_date}' AND '{end_date}'
```
evansd, line 80: I'd be tempted to remove the `standardise_if_date()` calls on L26-27 and use the [`quote()`](https://github.com/opensafely-core/cohort-extractor/blob/d25a266387003f8733f9256de5cfa37428ee4adc/cohortextractor/tpp_backend.py#L3617-L3628) function here instead e.g.
```
WHERE ConsultationDate BETWEEN {quote(start_date)} AND {quote(end_date)}
```

Same for the `population_query()` function below as well.

Building SQL via string interpolation is a bit yucky at the best of times, but I feel slightly better about it if the quoting is done right at the point where its used so you can't wonder whether it's been applied.

cohortextractor/generate_codelist_report.py
```diff
@@ -0,0 +1,109 @@
...
+    ) t
+    """
+
+
+def population_query(start_date, end_date):
+    # We're interested in patients who are registered on end_date, or who have died
+    # between start_date and end_date.
```
evansd, line 87: Given that we're counting coded events rather than patients it's not obvious to me why we need to worry much about the population at all. But I can well believe there's a good reason for it. Just wanted to check it was deliberate decision and hadn't been cargo-culted from elsewhere!

evansd (approved):
This looks great, and I think a strong validation of the hypothesis that it's significantly easier to do this with some custom SQL than trying to use the measures framework.

I think you could restructure things to make it even more efficient if you wanted to e.g. if you join on the population inside the `events_query()` then you can do the COUNT/GROUP BY in there as well and then you never have to build the list of individual events, which can be potentially quite large.

cohortextractor/generate_codelist_report.py
inglesp: This is deliberate, but thanks for checking!

lucyb: There's been quite a bit of discussion about the population used. So, at the request of the researchers, we're including only those patients mentioned. It might well change, but for now we're keeping it the same as the query we're comparing it with.

It might be relevant to the occurrence of codes because apparently, when someone leaves (or presumably joins) a practice there's an increase in codes, as they make sure their routine checks are done and any repeat prescriptions issued, etc, before they go. I suspect it probably evens out though!

evansd: Right yes, that makes sense. Thanks.

lucyb:
> This looks great, and I think a strong validation of the hypothesis that it's significantly easier to do this with some custom SQL than trying to use the measures framework.
> 
> I think you could restructure things to make it even more efficient if you wanted to e.g. if you join on the population inside the `events_query()` then you can do the COUNT/GROUP BY in there as well and then you never have to build the list of individual events, which can be potentially quite large.

That's a good point. I think it would be good to get this version tested first, but I definitely think you're right about there being room for further optimisation.

cohortextractor/generate_codelist_report.py
lucyb: Our strings have hyphens in them, which can apparently cause issues with SQL Server. The standardise_if_date function removes them.

Good point about the use of string interpolation though. pyodbc (which I assume we're using here), supports the use of parameterised queries, so I will give that a go

evansd: Oh sorry I should have said that the `quote()` function applies that date standardisation as well. (There's a whole [sorry story](https://github.com/opensafely-core/databuilder/issues/92) behind this that fortunately we have a proper solution to now in databuilder.)

Parameterised queries would be great (although they'll still need that date reformatting thing sadly).

evansd:
> I think it would be good to get this version tested first

Oh yes, I definitely meant that as a thing to worry about later. Or maybe not even at all if the current version turns out to be fast enough for our needs as it is.

cohortextractor/generate_codelist_report.py
lucyb: Oh, yes, sorry. I should've looked at the quote() function rather than making assumptions from the name. 

We don't seem to be making use of parameterised queries in the cohort extractor. Was there a reason for that or was it just faster+easier to use quote? I know they aren't suitable for all cases, so don't know if a standard approach was taken?

evansd: Originally we were generating SQL as text files which were fed into a separate Microsoft utility to actually execute them against the database, so parameterisation wasn't an option. This was before we managed to work around the issues involved with using pyODBC to fetch large amounts of data from MSSQL.

And yes, it's obviously incorrect for the `quote()` function to be doing the date transformation as well. But we only discovered the date inconsistency issue just before we were about to publish the first paper and so this was the easiest way of making sure we didn't miss any uses of dates anywhere.

There's a _lot_ of baggage from the early days in cohortextractor I'm afraid!

lucyb: Thank you. That makes a lot of sense. Code archaeology always brings up some interesting stuff.

I'm assuming the Microsoft utility was SQL Sever Agent? That brings back some memories!

PR #787, opened 2022-04-28
fix: Skip logging every INSERT line

rebkwok:


cohortextractor/log_utils.py
```diff
@@ -23,6 +23,7 @@
         structlog.processors.format_exc_info,
         structlog.processors.UnicodeDecoder(),
         structlog.stdlib.ProcessorFormatter.wrap_for_formatter,
+        structlog.processors.JSONRenderer(),
```
rebkwok, line 26: bah that shouldn't be there

cohortextractor/log_utils.py
```diff
@@ -74,8 +74,16 @@ def log_execution_time(logger, **log_kwargs):
...
+        first_non_comment_line = next(
+            line.strip() for line in sql_lines if not line.strip().startswith("--")
+        )
+        if first_non_comment_line.startswith("INSERT"):
+            # INSERTS can be lengthy, and are batched in 999s; logging all the lines here
+            # is not helpful and creates huge logs, so just log the first line for illustration
+            sql = first_non_comment_line + "..."
+        elif len(sql_lines) > 1000:
+            # Limit the SQL logged to a max 1000 lines
+            sql = "\n".join(sql_lines[:1000]) + "..."
```
bloodearnest, line 86: I think that putting a truncation marker on the end of a line might be confusing, maybe causing use to read that just that line was truncated, rather than lines after it?

Maybe have a truncation marker on it's own line at the end might be less ambiguous? 

If so, something like `[truncated]` might be clearer than `...`?

rebkwok: Done

bloodearnest (approved):

PR #788, opened 2022-04-28
fix: Log full SQL on the first index date only

rebkwok:
If there is a range of index dates, the executed SQL is identical for each pass, with the exception of the index date string.  Logging the full SQL each time isn't informative and just inflates the logs.  Instead, log it in full for the first index date, and truncate it to the first line for subsequent index dates, which should still allow us to identify the specific SQL executed when a job failed or was particularly slow, even if we have to look back to the corresponding line for the first index date to find it.

### DIFF
--- a/cohortextractor/cohortextractor.py
+++ b/cohortextractor/cohortextractor.py
@@ -198,6 +198,10 @@ def _generate_cohort(
     if index_date_range:
         log_stats(logger, min_index_date=index_dates[-1], max_index_date=index_dates[0])
 
+    # record if we've logged the SQL for this generate_cohort run
+    # For subsequent runs, we log the timing and the first line of the SQL only, since
+    # the only difference should be the index date
+    sql_logged = False
     for index_date in index_dates:
         log_event = f"generate_cohort for {study_name}"
         if index_date is not None:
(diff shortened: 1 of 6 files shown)


cohortextractor/cohortextractor.py
```diff
@@ -206,6 +210,8 @@ def _generate_cohort(
             if index_date is not None:
                 logger.info(f"Setting index_date to {index_date}")
                 study.set_index_date(index_date)
+                if sql_logged:
+                    study.set_sql_logging(truncate=True)
```
bloodearnest, line 214: What is the scope of `study` here? 

It it possible the same `study` object will be later reused for non-index date queries, and thus truncate them when we didn't intend to?

bloodearnest (approved):
This looks good, tests are great and the comments super helpful.

Left one question, which I expect is fine, but just double checking.

cohortextractor/cohortextractor.py
rebkwok: No, it shouldn't.  It gets instantiated in this function, before the index dates loop, with `load_study_definition`, at which point it has the default truncate=False, and the function doesn't return anything after the loop

PR #789, opened 2022-04-29
Update generate codelist report

inglesp:
In particular, we now:

* break down counts by week and practice, instead of by week
* compute the list size for each practice

This lets us compute deciles over all practices of codelist use per week.

We also now:

* only count one event per patient per day


tests/test_generate_codelist_report.py
```diff
@@ -150,44 +146,31 @@ def create_patients():
...
-    dead_patient.CodedEventsSnomed = [
-        CodedEventSnomed(ConceptID=code, ConsultationDate=date)
-        for code, date in dead_patient_events
-    ]
+    session = make_session()
+    session.add(current_patient)
+    session.add(dead_patient)
+    session.add(former_patient)
+    session.add(long_dead_patient)
+    session.commit()
```
inglesp, line 165: Move this to setup() function.

cohortextractor/generate_codelist_report.py
```diff
@@ -83,6 +91,31 @@ def events_query(start_date, end_date):
...
+        FROM (
+            SELECT RegistrationHistory.Patient_ID, Organisation_ID,
+            ROW_NUMBER() OVER (
+                PARTITION BY RegistrationHistory.Patient_ID
+                ORDER BY StartDate DESC, EndDate DESC, Registration_ID
+            ) AS rownum
+            FROM RegistrationHistory
+            WHERE StartDate <= {end_date}
+        ) t
+        LEFT JOIN Organisation
```
inglesp, line 144: Do we need to join on organisation?

tests/test_generate_codelist_report.py
```diff
@@ -89,22 +82,28 @@ def test_generate_codelist_report(tmpdir):
         ("2022-01-26", "22222222"),
     ]
 
-    #
-    expected_counts_per_code = {"11111111": 36, "22222222": 6}
+    expected_counts_per_code = {"11111111": 24, "22222222": 6}
     expected_counts_per_week = {
```
inglesp, line 87: Rename to counts_per_practice_per_week?

cohortextractor/generate_codelist_report.py
```diff
@@ -40,10 +41,17 @@ def generate_codelist_report(output_dir, codelist_path, start_date, end_date):
...
-    counts = pd.DataFrame(list(cursor), columns=["code", "date", "num"])
+    counts = pd.DataFrame(list(cursor), columns=["code", "date", "practice", "num"])
 
     counts_per_code = counts.groupby("code")["num"].sum()
-    counts_per_week = counts.groupby(pd.Grouper(key="date", freq="W-MON"))["num"].sum()
+
+    # Computing counts per practice per week  is slightly more involved than computing
+    # counts per code, because we need to account for weeks when a matching code is not
+    # recorded at a practice.
+    grouper = pd.Grouper(key="date", freq="W-MON")
```
inglesp, line 57: Call it `week_grouper`?

cohortextractor/generate_codelist_report.py
lucyb: Realised my confusion earlier was mostly because I hadn't seen the `WHERE t.rownum = 1`, so removing the join on organisation will make that much more obvious.

Is that being covered by the tests at the moment?

lucyb: Just to clarify my understanding, the reason for the practice query is that a patient may have been registered at more than one practice when we do `StartDate <= {end_date} AND EndDate > {end_date}` in `population_query()`. So the purpose of `practice_query()` is to get the most recent practice.

Would it be worth renaming `practice_query` to `most_recent_practice_query`, or if that's too much of a mouthful, to add a code comment to say why this is needed?

lucyb (approved):

PR #791, opened 2022-05-03
Exit codes for specific database errors

### LINKED ISSUE #776: Use exit codes to indicate specific error conditions
Related to this job-runner issue:
 * https://github.com/opensafely-core/job-runner/issues/184#issuecomment-1090213792

We should add another except block here:
https://github.com/opensafely-core/cohort-extractor/blob/acf8508e25a39493a7a7beae861e57f2c4db7a43/cohortextractor/cohortextractor.py#L684-L686

This should catch all exceptions from the database (exactly what class or classes this involves is an open question) and:
 * print the trackback in the normal way by calling [`traceback.print_exc()`](https://docs.python.org/3/library/traceback.html#traceback.print_exc);
 * inspect the contents of the exception to determine the appropriate error code;
 * call `sys.exit()` with that code.

The three cases we want to distinguish at the moment are:
 * `Unexpected EOF from the server` — [example job](https://jobs.opensafely.org/datalab/brit-antibiotic-research/hosp_pred/6402/thdai4ntokyzgwfo/)
 * "Missing CodedEvent_SNOMED table" (I don't have the exact text for this, will need digging out the job log) — [example job](https://jobs.opensafely.org/university-of-leeds/cancer-surgery-service-evaluation/all_cause_mortality_with_stratification/6577/x5tvqrb4lpuiapam/)
 * Any other sort of database error

We'll need to choose exit codes for this which don't clash with anything we might be likely to generate otherwise. See:
https://tldp.org/LDP/abs/html/exitcodes.html
https://www.freebsd.org/cgi/man.cgi?query=sysexits&apropos=0&sektion=0&manpath=FreeBSD+4.3-RELEASE&format=html

rebkwok:
fixes #776 

Catches [pymssql errors](https://github.com/pymssql/pymssql/blob/f486f7d79bd2583eba96323cdeca34eb8f14cd2a/src/pymssql/_pymssql.pyx#L117); of the ones listed there, `OperationalError`, `ProgrammingError` and `IntegrityError` seem to be the ones actually raised by pymssql.  

The 2 specific errors from issue #776 are: 

- `Unexpected EOF from the server` (`OperationalError`)
- `Invalid object name 'CodedEvent_SNOMED'` when the snomed table is being built (`ProgrammingError`).  

Other database errors should raise one of these, and I've included `IntegrityError` just in case so that would also get caught as a general database error code.

### DIFF
--- a/cohortextractor/cohortextractor.py
+++ b/cohortextractor/cohortextractor.py
@@ -774,24 +774,27 @@ def main():
             print(f"Dummy data error: {e}")
             sys.exit(1)
         except Exception as e:
-            traceback.print_exc()
             # Checking for "DatabaseError" in the MRO means we can identify database errors without
             # referencing a specific driver.  Both pymssql and presto/trino-python-client raise
             # exceptions derived from a DatabaseError parent class
-            if "DatabaseError" in str(e.__class__.mro()):
-                # Exit with specific exit codes to help identify known issues
-                if "Unexpected EOF from the server" in str(e):
-                    logger.error(f"Intermittent database error: {e}")
-                    sys.exit(3)
-                if "Invalid object name 'CodedEvent_SNOMED'" in str(e):
-                    logger.error(
-                        "CodedEvent_SNOMED table is currently not available.\n"
-                        "This is likely due to regular database maintenance."
-                    )
-                    sys.exit(4)
-                logger.error(f"Database error: {e}")
-                sys.exit(5)
-            sys.exit(1)
+
+            # Ignore errors which don't look like database errors
+            if "DatabaseError" not in str(e.__class__.mro()):
+                raise
+
+            traceback.print_exc()
+            # Exit with specific exit codes to help identify known issues
+            if "Unexpected EOF from the server" in str(e):
+                logger.error(f"Intermittent database error: {e}")
+                sys.exit(3)
+            if "Invalid object name 'CodedEvent_SNOMED'" in str(e):
+                logger.error(
+                    "CodedEvent_SNOMED table is currently not available.\n"
+                    "This is likely due to regular database maintenance."
+                )
+                sys.exit(4)
+            logger.error(f"Database error: {e}")
+            sys.exit(5)
 
     elif options.which == "generate_measures":
         generate_measures(
--- a/cohortextractor/cohortextractor.py
+++ b/cohortextractor/cohortextractor.py
@@ -35,7 +35,6 @@ from cohortextractor.exceptions import DummyDataValidationError
 from cohortextractor.generate_codelist_report import generate_codelist_report
 
 from .log_utils import log_execution_time, log_stats
-from .mssql_utils import database_exceptions
 
 logger = structlog.get_logger()
 
@@ -774,20 +773,25 @@ def main():
         except DummyDataValidationError as e:
             print(f"Dummy data error: {e}")
             sys.exit(1)
-        except database_exceptions() as e:
+        except Exception as e:
             traceback.print_exc()
-            # Exit with specific exit codes to help identify known issues
-            if "Unexpected EOF from the server" in str(e):
-                logger.error(f"Intermittent database error: {e}")
-                sys.exit(151)
-            if "Invalid object name 'CodedEvent_SNOMED'" in str(e):
-                logger.error(
-                    "CodedEvent_SNOMED table is currently not available.\n"
-                    "This is likely due to regular database maintenance."
-                )
-                sys.exit(152)
-            logger.error(f"Database error: {e}")
-            sys.exit(153)
+            # Checking for "DatabaseError" in the MRO means we can identify database errors without
+            # referencing a specific driver.  Both pymssql and presto/trino-python-client raise
+            # exceptions derived from a DatabaseError parent class
+            if "DatabaseError" in str(e.__class__.mro()):
+                # Exit with specific exit codes to help identify known issues
+                if "Unexpected EOF from the server" in str(e):
+                    logger.error(f"Intermittent database error: {e}")
+                    sys.exit(3)
+                if "Invalid object name 'CodedEvent_SNOMED'" in str(e):
+                    logger.error(
+                        "CodedEvent_SNOMED table is currently not available.\n"
+                        "This is likely due to regular database maintenance."
+                    )
+                    sys.exit(4)
+                logger.error(f"Database error: {e}")
+                sys.exit(5)
+            sys.exit(1)
 
     elif options.which == "generate_measures":
         generate_measures(
--- a/cohortextractor/cohortextractor.py
+++ b/cohortextractor/cohortextractor.py
@@ -12,6 +12,7 @@ import os
 import pathlib
 import re
 import sys
+import traceback
 from argparse import ArgumentParser
 from collections import defaultdict
 from io import BytesIO
@@ -34,6 +35,7 @@ from cohortextractor.exceptions import DummyDataValidationError
 from cohortextractor.generate_codelist_report import generate_codelist_report
 
 from .log_utils import log_execution_time, log_stats
+from .mssql_utils import database_exceptions
 
 logger = structlog.get_logger()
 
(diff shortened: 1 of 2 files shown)


cohortextractor/cohortextractor.py
```diff
@@ -772,6 +774,21 @@ def main():
...
+            if "Unexpected EOF from the server" in str(e):
+                logger.error(f"Intermittent database error: {e}")
+                sys.exit(151)
+            if "Invalid object name 'CodedEvent_SNOMED'" in str(e):
+                logger.error(
+                    "CodedEvent_SNOMED table is currently not available.\n"
+                    "This is likely due to regular database maintenance."
+                )
+                sys.exit(152)
+            logger.error(f"Database error: {e}")
```
rebkwok, line 789: I've made these log here, but possibly job-runner should be responsible for looking for the specifc exit codes and logging there instead?  It checks for 137 (likely memory issues), but only in the local executor afaict.

bloodearnest: I think handling them here is correct.

In general, the sooner you handle an exception, the more context you have.

The job-runner only checks for 137 specifically because that very likely means the process was killed by the OOM, so it had no chance internally to catch and handle the error.

cohortextractor/cohortextractor.py
```diff
@@ -772,6 +774,21 @@ def main():
         except DummyDataValidationError as e:
             print(f"Dummy data error: {e}")
             sys.exit(1)
+        except database_exceptions() as e:
+            traceback.print_exc()
+            # Exit with specific exit codes to help identify known issues
+            if "Unexpected EOF from the server" in str(e):
+                logger.error(f"Intermittent database error: {e}")
+                sys.exit(151)
```
bloodearnest, line 782: I don't think we should be using and exit code above 128, unless there's something I'm missing?

This is our process, we control the code, so we can decide what its exit codes mean.

So I would suggest we use 2, 3, 4, etc, for different error cases (maybe reserving exit code of 1 for unhandled errors).

Error codes above 128 typically mean that the process was killed externally by a signal, the exit code being 128 + signal number. So, and exit code 137 means the process was killed by signal number 9 (SIGKILL), of `kill -9` fame, which typically means the OOM killer, which users the most powerful signal available.

cohortextractor/cohortextractor.py
```diff
@@ -772,6 +774,21 @@ def main():
         except DummyDataValidationError as e:
             print(f"Dummy data error: {e}")
             sys.exit(1)
+        except database_exceptions() as e:
```
bloodearnest, line 777: This is interesting, I've never seen except clauses dynamically generated before, but now I think of it, it seems plausible.

I did find this surprising, though. Would this work if we do the try/catch at the module level, and define this list at import time, rather than re-checking every time?

rebkwok: So it's like this because otherwise a test fails if you import pymssql at the module level.  I think it'll still complain if the try/catch is at the modue level, because it'll have pymssql in sys.modules

bloodearnest: Could we cache the result of the function at module level, but leave the import inside the function scope?

e.g.
```
DATABASE_EXCEPTIONS = database_exceptions()
```
then
```
        except DATABASE_EXCEPTIONS as e:
```

cohortextractor/mssql_utils.py
```diff
@@ -194,3 +193,23 @@ def fetch(self, min_key=None):
...
+    try:
+        import pymssql
+
+        return (
+            pymssql._pymssql.OperationalError,
+            pymssql._pymssql.ProgrammingError,
+            pymssql._pymssql.IntegrityError,
+        )
+    except ImportError:
+        return ()
```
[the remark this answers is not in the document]
rebkwok: I guess really it should check the other drivers for database-related errors, but given that we only use pymssql now, it didn't seem worth doing, and my intention was that it woulddri return something that wouldn't every get caught, and db errors raised by other drivers (which presumably wouldn't happen anymore unless someone had swapped in another one for some reason) would get raised normally and exit with the default exit code 1.  I don't know that we want to catch all exceptions here though, that would make it always raise a general database error.  It could return a dummy exception class instead?

cohortextractor/cohortextractor.py
bloodearnest: yeah, max signal numbers and exit codes conventions vary a lot, so being conservative is a good idea, and that seems a reasonable range.

Running cohortextractor jobs in a BSD jail sounds like my weekend plans now :D

I've been assuming that the main goal of using different exit codes here is so that the job-runner executor can communicate something more useful to the user about the specifics of why their job failed. We could in theory have some convention of showing the last line of log output, or similar, which would mean any non-zero code would be fine. But then the old privacy leak issue rears its head, so I guess we need to communicate via ERRNOs like its 1989.

It would be really great if there was a way to safely communicate a short string from the docker process. I guess we could check for specific strings on the last few lines, which is almost as bad as semantic exit codes, but more flexible.

rebkwok: I've updated this PR to look for the string "DatabaseError" instead of specific exceptions, and to use exit codes within the available range.  Couldn't quite figure out from the last couple of comments here if either of you were suggesting we should ditch this approach altogether.

evansd: Thanks Becky, will take a look.

I was agreeing with @bloodearnest that this _is_ all a bit 1989, but also trying to argue that it's our best option at the moment. So unless Simon's got a brilliant leftfield idea up his sleeve (not a possibility to discount!) I think we should go ahead.

rebkwok: Given that cohort-extractor's going away at some point, I think generally doing the easiest thing at the moment is probably best

bloodearnest: Sorry, was musing about an alternative, but @dave's point about executor API changes is valid, and I think exit codes is the only way for now, sorry for the confusion.

cohortextractor/cohortextractor.py
```diff
@@ -774,20 +773,25 @@ def main():
         except DummyDataValidationError as e:
             print(f"Dummy data error: {e}")
             sys.exit(1)
-        except database_exceptions() as e:
+        except Exception as e:
             traceback.print_exc()
```
evansd, line 777: I think it might be cleaner to just reraise non-database exceptions rather than try to handle them ourselves. So something like:
```py
if "DatabaseError" in str(e.__class__.mro()):
    traceback.print_exc()
    ...
    sys.exit(5)
else:
    raise
```

Or maybe even an early exit to avoid the indentation:
```py
# Ignore errors which don't look like database errors
if "DatabaseError" not in str(e.__class__.mro()):
    raise

traceback.print_exc()
sys.exit(5)
```

evansd (approved):

cohortextractor/cohortextractor.py
rebkwok: That makes sense.  Does calling `raise` mean that it will exit with code 1 anyway?  I followed what was done for the `DummyDataValidation` block, but wondering if that should just call `raise` too?

evansd: Yes, behaviourally it will be basically the same as what you already have: it will print a traceback and exit 1. So you definitely don't _have_ to make this change. 

The difference with the `DummyDataValidation` block is that it's only catching one class of exception, and we deliberately don't want to print the full trackback there — just the error itself — to avoid confusing users. So I think that can stay as it is.

The pattern I was suggesting here is effectively trying to do the same thing as the `DummyDataValidation` block. But, because we can't statically define the class of exceptions we want to catch, we have to catch everything and then re-raise if it turns out to be not the sort of exception we wanted to catch. (A bare `raise` statement is effectively saying "pretend I never caught this exception in the first place".)

Does that make sense?

rebkwok: It does, thanks :slightly_smiling_face: I've made that change now too.

evansd (approved):
:+1:

PR #792, opened 2022-05-06
fix: Log start, stop and state for timing logs

rebkwok:


cohortextractor/cohortextractor.py
```diff
@@ -204,10 +204,10 @@ def _generate_cohort(
     # the only difference should be the index date
     sql_logged = False
     for index_date in index_dates:
-        log_event = f"generate_cohort for {study_name}"
+        log_kwargs = dict(description="generate_cohort", study=study_name)
```
bloodearnest, line 207: Does "study" mean "study_defintion" here? If so possibly worth being explicit?

cohortextractor/log_utils.py
```diff
@@ -71,29 +71,53 @@ def log_stats(logger, **kwargs):
...
+        log_kwargs["state"] = "ok"
     finally:
-        elapsed_time = process_time() - start
+        stop = process_time()
+        if sql and not sql.endswith("[truncated]") and len(sql_lines) > 2:
+            # if applicable we logged the full SQL at the start of the timing block
+            # Log just the truncated first line here at the end to help match up logs
+            # Only truncate if the total lines is > 2; this will keep a comment line
+            # with a single line of SQL
+            log_kwargs["sql"] = _truncate([first_non_comment_line])
```
bloodearnest, line 116: I wonder if it would be better to do this via numbering queries? And then log start of query N and the end of query N? That's more explicit than matching sql, and less noisy too?

rebkwok: What do you mean by numbering queries?  We have all the queries that get built by the backend and executed in a loop, but there are others that happen ouside of that I think (e.g. the ones to write in batches to a temporary file).  I thought about assigning a short uuid instead/as well as the first line(s) of sql - would that work better?

bloodearnest: if we have a global counter at the module level, e.g.

```python
query_counter = itertools.count()
```

Then our log_execution_time() can start with:

```python

    query_id = next(query_counter)
    log_stats["query_id"] = query_id
```

That's much more readable and less noisy than uuid IMO. And it serves the purpose of uniquely identifying queries within a single run.

bloodearnest (approved):

PR #799, opened 2022-05-11
Choose bucket day from start date

inglesp:
Fixes opensafely-core/interactive.opensafely.org#95

To the reviewer: please could you independently check that the comments in test data setup are correct, and that `expected_counts_per_week_per_practice` is also correct.


iaindillingham:
The comments in `event_data` are correct in the sense that the reasons given for including/not including correspond to the values of `expected_counts_per_week_per_practice`.

I'm not sure if this is helpful, but I don't think you've tested that the grouper is left-closed/right open (`lower >= value < upper`).

inglesp:
> I'm not sure if this is helpful, but I don't think you've tested that the grouper is left-closed/right open (`lower >= value < upper`).

The test fails if we change

```
    week_grouper = pd.Grouper(key="date", label="left", closed="left", freq=freq)
```

to

```
    week_grouper = pd.Grouper(key="date", label="left", closed="right", freq=freq)
```

Is that what you meant?

iaindillingham (approved):

lucyb (approved):

PR #802, opened 2022-05-12
feat: add flag to turn on correct end date handling

inglesp:


tests/tpp_backend_setup.py
```diff
@@ -25,6 +26,25 @@
 Base = declarative_base()
 
 
+# Hack: in order to improve the variety of our test data, this mixin intercepts object
+# initialization, adding a time part to any kwarg that is for a datetime field that has
+# been provided as a date.
+#
+# (We cannot add this behaviour by subclassing Base, because subclasses of Base are
+# expected to have a __tablename__ attribute.)
```
evansd, line 34: You can use the `cls` argument to `declarative_base()` to set the base class:
https://docs.sqlalchemy.org/en/13/orm/extensions/declarative/api.html#sqlalchemy.ext.declarative.declarative_base

cohortextractor/cohortextractor.py
```diff
@@ -31,6 +31,7 @@
 )
 
 import cohortextractor
+from cohortextractor import flags
```
evansd, line 34: I think you forgot to `git add` the module.

cohortextractor/cohortextractor.py
```diff
@@ -685,6 +686,10 @@ def main():
         help="Do not regenerate data if output file already exists",
         action="store_true",
     )
+    generate_cohort_parser.add_argument(
+        "--with-end-date-fix",
+        action="store_true",
```
evansd, line 691: I'd be tempted to reverse the logic here. Don't we want it fixed by default, and then separately we want to be able to re-enable the bug to do some investigation as to its impact. Is there value to keeping the bug in place?

evansd (changes requested):
I think this looks basically correct, modulo my comments below.

cohortextractor/cohortextractor.py
inglesp: I had a comment explaining this in the missing flags.py.  I've moved that comment into cohortextractor.py -- what do you think?

evansd: Yes, that seems reasonable

evansd (approved):

PR #811, opened 2022-06-08
Pass DATABASE_URL to docker-compose

iaindillingham:
This should enable us to call cohortextractor dump_cohort_sql without error.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -22,4 +22,4 @@ WORKSPACE=${WORKSPACE:-$PWD}
 export DOCKER_BUILDKIT=1
 export COMPOSE_DOCKER_CLI_BUILD=1
 
-exec docker-compose run --rm -v "$WORKSPACE:/workspace" "$ENV" "$@"
+exec docker-compose run -e DATABASE_URL --rm -v "$WORKSPACE:/workspace" "$ENV" "$@"


StevenMaude:
This is doing what it is supposed to.

Looking again at how [Docker Compose picks up environment variables](https://docs.docker.com/compose/environment-variables/), it's worth noting that you can set those variables several ways (in `docker-compose.yml`, `.env`, in the shell…, at the Docker command line…)

Which gives a lot of control but is also potentially very confusing.

It _might_ cause confusion if you have an existing `.env` file, and expect those variables to be used, but forget you've set `DATABASE_URL` in your shell environment somewhere, temporarily or permanently. **The shell variable takes precedence.** Worth noting that the variable name is also fairly generic here.

The alternative solution here would be just to edit the `.env` file directly. I don't know if there's added convenience of setting the URL at shell level, instead of editing a line or renaming different configurations to `.env` to change it. If setting at shell level is useful, then this is OK.

iaindillingham:
I'm going to close this PR without merging, because I think the motivation is no longer motivating enough 🙂 

For archaeologists: @milanwiedemann wanted to test a study definition's logic. I suggested dumping the SQL; reading the SQL might give him more insight into the study definition's logic.

I ran

```
WORKSPACE=path/to/study ./run.sh cohortextractor dump_cohort_sql
```

but this raised

```
RuntimeError: Cannot extract data as no DATABASE_URL environment variable defined
```

so I ran

```
DATABASE_URL=mssql:// WORKSPACE=path/to/study ./run.sh cohortextractor dump_cohort_sql
```

but this raised the same error. This PR passes `DATABASE_URL` to the container. As @StevenMaude says, though, this could be confusing.

What I didn't appreciate then, but do now, is that to execute the above, we have to clone cohort-extractor. And having cloned cohort-extractor, we could create a virtual environment. And with a virtual environment, we don't need to run `./run.sh`. Instead, we can run

```
DATABASE_URL=mssql:// WORKSPACE=path/to/study cohortextractor dump_cohort_sql
```

and we're done 🛑

PR #812, opened 2022-06-09
SQL server stats timing

rebkwok:
Have SQL Server include timings (parse/compile and execute) for each query:
https://docs.microsoft.com/en-us/sql/t-sql/statements/set-statistics-time-transact-sql

Note that I can't find a way to tell what the related query is when logging the timing message.  We have a timing id for all timed queries, and I'm tagging the log with that, but that assumes that the timing message relates to the last SQL query that went through the `cursor.execute` call.  I don't think that will always be the case - if parse/compile happens when `cursor.execute()` is called, but actual execution doesn't happen until results are fetched, then execution timing could be delayed until after other queries have been compiled.  In that case, we'll see duplicate logs for execution time with the same timing id,  and that should happen for the last/near to last id in a generate_cohort run.  From the investigation I was able to do with tests, I think this may only happen once in a single generate_cohort run (there could be multiple runs if there's an index date range) and we may be able to extrapolate by looking at which timing ids only have a parse/compile timing logged. 

WIth that caveat, I think it'll still be interesting to run this on real jobs and see if there are any interesting patterns that show up.  Hopefully we can replace this with more explicit explorations with [DMV](https://docs.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-exec-query-stats-transact-sql?view=sql-server-ver16), when(if) we get the required permissions to use it.

### DIFF
--- a/cohortextractor/mssql_utils.py
+++ b/cohortextractor/mssql_utils.py
@@ -5,7 +5,7 @@ from urllib.parse import unquote, urlparse
 
 import structlog
 
-from .log_utils import log_execution_time
+from .log_utils import log_execution_time, timing_log_counter
 
 # Some drivers warn about the use of features marked "optional" in the DB-ABI
 # spec, using a standardised set of warnings. See:
@@ -93,6 +93,32 @@ def _ctds_connect(ctds, params):
     return ctds.connect(**params)
 
 
+def stats_msg_handler(msgstate, severity, srvname, procname, line, msgtext):
+    """
+    Log parse, compile and execution timing messages sent by the server.
+    """
+    timing_re = re.compile(
+        r"""
+        .*SQL\sServer\s(?P<timing_type>parse\sand\scompile\stime|Execution\sTime).*  # SQl Server Statistics time message prefix
+        .*CPU\stime\s=\s(?P<cpu_time>\d+)\sms  # cpu time in ms
+        .*elapsed\stime\s=\s(?P<elapsed_time>\d+)\sms  # elapsed time in ms
+        """,
+        flags=re.S | re.X,
+    )
+    if timing_match := timing_re.match(msgtext.decode()):
+        logger.info(
+            "sqlserver-stats",
+            description=timing_match.group("timing_type").lower().replace(" ", "_"),
+            cpu_time_secs=int(timing_match.group("cpu_time")) / 1000,
+            elapsed_time_secs=int(timing_match.group("elapsed_time")) / 1000,
+            # SQL Server statistics timing is only set to ON for timed query execution, so
+            # we may be able match it to the query it timed by tagging it with the current timing id
+            # This may result in multiple execution timings logged against one timing id if
+            # query execution is delayed until the time of fetching
+            timing_id=timing_log_counter.current,
+        )
+
+
 def _pymssql_connect(pymssql, params):
     # https://pymssql.readthedocs.io/en/stable/ref/pymssql.html#pymssql.connect
     params = params.copy()
@@ -102,7 +128,11 @@ def _pymssql_connect(pymssql, params):
     # set to one week
     params["timeout"] = 7 * 24 * 60 * 60
     params["autocommit"] = True
-    return pymssql.connect(**params)
+
+    connection = pymssql.connect(**params)
+    # Install our custom stats handler
+    connection._conn.set_msghandler(stats_msg_handler)
+    return connection
 
 
 def mssql_fetch_table(
--- a/cohortextractor/mssql_utils.py
+++ b/cohortextractor/mssql_utils.py
@@ -14,6 +14,15 @@ warnings.filterwarnings("ignore", re.escape("DB-API extension cursor.__iter__() 
 
 logger = structlog.get_logger()
 
+SQLSERVER_TIMING_REGEX = re.compile(
+    r"""
+        .*SQL\sServer\s(?P<timing_type>parse\sand\scompile\stime|Execution\sTime).*  # SQl Server Statistics time message prefix
+        .*CPU\stime\s=\s(?P<cpu_time>\d+)\sms  # cpu time in ms
+        .*elapsed\stime\s=\s(?P<elapsed_time>\d+)\sms  # elapsed time in ms
+        """,
+    flags=re.S | re.X,
+)
+
 
 def mssql_connection_params_from_url(url):
     parsed = urlparse(url)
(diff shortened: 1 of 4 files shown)


evansd:
> if parse/compile happens when cursor.execute() is called, but actual execution doesn't happen until results are fetched, then execution timing could be delayed until after other queries have been compiled

My (reasonably confident) guess is that `execute()` is actually completely synchronous and it was just the misleading timings we got by using `process_time()` that led us to think otherwise.

rebkwok:
> My (reasonably confident) guess is that `execute()` is actually completely synchronous and it was just the misleading timings we got by using `process_time()` that led us to think otherwise.

I'm not sure that's the case - I'm definitely seeing some oddness in the tests.  Some execute queries only log a parse/compile step, some only log an execution step (maybe there isn't a parse/compile for some things like creating indexes?), and the timing ID for the DROP TABLE query at the end logs execution twice

rebkwok:
Oh wait...writing out the queries using `mssql_fetch_table` returns a generator; I wonder if it's just that messing with what we see as the "last" timing ID in the message handler

rebkwok:
If that's the case, I can't think of a way to get it to log the right timing id for the download queries (the `SELECT TOP 32000...` queries), but we should only end up with multiple execution timings for one ID, the DROP TABLE one, of which the last one will actually be the DROP TABLE query, and the rest will be the download queries

rebkwok:
Added to the tests to verify that this does indeed seem to be what's happening

cohortextractor/mssql_utils.py
```diff
@@ -93,6 +102,24 @@ def _ctds_connect(ctds, params):
     return ctds.connect(**params)
 
 
+def stats_msg_handler(msgstate, severity, srvname, procname, line, msgtext):
+    """
+    Log parse, compile and execution timing messages sent by the server.
+    """
+    if timing_match := SQLSERVER_TIMING_REGEX.match(msgtext.decode()):
```
bloodearnest, line 109: image

cohortextractor/mssql_utils.py
```diff
@@ -93,6 +102,24 @@ def _ctds_connect(ctds, params):
...
+        logger.info(
+            "sqlserver-stats",
+            description=timing_match.group("timing_type").lower().replace(" ", "_"),
+            cpu_time_secs=int(timing_match.group("cpu_time")) / 1000,
```
bloodearnest, line 113: suggestion: add units to the name, so "cpu_time_s" or "cpu_time_ms" so it obvious to the reader of logs straight away

cohortextractor/mssql_utils.py
```diff
@@ -14,6 +14,15 @@
...
 
+SQLSERVER_TIMING_REGEX = re.compile(
+    r"""
+        .*SQL\sServer\s(?P<timing_type>parse\sand\scompile\stime|Execution\sTime).*  # SQl Server Statistics time message prefix
+        .*CPU\stime\s=\s(?P<cpu_time>\d+)\sms  # cpu time in ms
+        .*elapsed\stime\s=\s(?P<elapsed_time>\d+)\sms  # elapsed time in ms
+        """,
+    flags=re.S | re.X,
+)
+
```
bloodearnest, line 25: It is possible to get the statistics as XML. Not sure if thats better than regexing plain text though!

cohortextractor/mssql_utils.py
```diff
@@ -93,6 +102,24 @@ def _ctds_connect(ctds, params):
     return ctds.connect(**params)
 
 
+def stats_msg_handler(msgstate, severity, srvname, procname, line, msgtext):
```
bloodearnest, line 105: I think this handler should probably have a big `try: ... except Exception: pass` around the body. We don't want any errors in stats parsing to stop the job from running.

tests/test_logging.py
```diff
@@ -817,14 +660,75 @@ def assert_stats_logs(
...
+    # one SQL Server parse/compile log and one execution log;
+    # **EXCEPT** CREATE INDEX and DROP TABLE queries, which don't create a parse/compile
+    # timing, only an execution timing
+
+    # However, when we generate the queries to select the final results and write
+    # them to file, we return a generator.  At the point where the query is actually
+    # executed and the timing message handler is called, the last timing id is the one for
+    # dropping the temporary results table query. So we get no execution_time log for the
+    # download queries' timing_ids, and the DROP TABLE query's timing id will collect execution timing for
+    # both the download queries and (finally) the real drop table query
```
bloodearnest, line 686: arg, this is a bit of a mess isn't it.

I suspect there's a better way to do this tracking, but it escapes me on a Friday afternoon.

bloodearnest (approved):
LGTM.

A suggestion on log variable naming, and some exception handling to be cautious.

cohortextractor/mssql_utils.py
rebkwok: it's already got units in the name? `cpu_time_secs` is what we'll see in logs

cohortextractor/mssql_utils.py
rebkwok: First time actually using it :)

cohortextractor/mssql_utils.py
rebkwok: Is it?  How do I do that?  Any idea if that would also get me the query it relates to?

rebkwok: Also not sure if it's any better/easier!

tests/test_logging.py
rebkwok: I'm sure there must be too, but I couldn't find one

cohortextractor/mssql_utils.py
bloodearnest: doh, sorry! perfect

cohortextractor/mssql_utils.py
bloodearnest: `SET STATISTICS XML ON`, possibly. Not sure its worth it. https://docs.microsoft.com/en-us/sql/t-sql/statements/set-statistics-xml-transact-sql?view=sql-server-ver16

It might also be a completely separate thing from `SET STATISTICS TIME ON`, so not sure if its worth pursuing

rebkwok: I think it would be worth it - it looks like it might give us more details of the query - but I don't think there's a way to get it out of pymssql easily

PR #813, opened 2022-06-10
Add urban_rural_classification lookup to docs

alexwalkerepi:
This is the mapping for how this variable is encoded in TPP.

I'm unsure if this is the right place for this to be documented. While the classification is commonly used elsewhere, it's possible other (future) data sources might encode it differently. On the other hand, we need it to be somewhere discoverable to people using the variable.

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -1157,7 +1157,15 @@ def address_as_of(
         returning: string indicating value to be returned. Options are:
 
             * `index_of_multiple_deprivation`
-            * `rural_urban_classification`
+            * `rural_urban_classification` encoded (in at least TPP) as:
+                * 1 - Urban major conurbation
+                * 2 - Urban minor conurbation
+                * 3 - Urban city and town
+                * 4 - Urban city and town in a sparse setting
+                * 5 - Rural town and fringe
+                * 6 - Rural town and fringe in a sparse setting
+                * 7 - Rural village and dispersed
+                * 8 - Rural village and dispersed in a sparse setting
             * `msoa`
 
         round_to_nearest: an integer that represents how `index_of_multiple_deprivation` value are rounded.


cohortextractor/patients.py
```diff
@@ -1165,9 +1165,19 @@ def address_as_of(
         return_expectations: a dict defining the `rate` and the `categories` returned with ratios
 
     Returns:
-        list: of integers for `rural_urban_classification` and `index_of_multiple_deprivation`, strings for `msoa`.
-            (IMD is a ranking from 1 to around 33000 (the number of LSOAs in England), where 1 represents most
-        deprived.)
+        list: of integers for and `index_of_multiple_deprivation`, strings for `msoa`.
```
rebkwok, line 1168: There's a missing `rural_urban_classification` here

cohortextractor/patients.py
```diff
@@ -1165,9 +1165,19 @@ def address_as_of(
...
+            `rural_urban_classification` is encoded (in at least TPP) as:
+                * 1 - Urban major conurbation
+                * 2 - Urban minor conurbation
+                * 3 - Urban city and town
+                * 4 - Urban city and town in a sparse setting
+                * 5 - Rural town and fringe
+                * 6 - Rural town and fringe in a sparse setting
+                * 7 - Rural village and dispersed
+                * 8 - Rural village and dispersed in a sparse setting
+            `index_of_multiple_deprivation` (IMD) is a ranking from 1 to around 33000 (the number of LSOAs
```
rebkwok, line 1178: While you're here, can you update "around 33000" to "32800" for consistency with other changes that Iain's made recently to the IMD docs

rebkwok (changes requested):
I think this is the right place for it :+1: 

At least one of the commits needs to be a semantic commit (prefixed with "feat:" or "fix:") so that a new package will be built.  Once the new package is build, the documentation repo can be updated with the new version, which will pull the changes through to the main docs.

rebkwok (approved):

PR #815, opened 2022-06-15
feat: support dummy data for generate_codelist_report command

bloodearnest:
This command does not use study definitions, so we use previously
developed code to embed a default dummy data model for it. This allows
it to be used in local development. Currently, it can only be
run in production.

Note: this work uncovered a bug in the db fixutre cleanup, in that
removing the tables didn't actually work once there was multiple tests
in the module. Rather than figure out the specific deletion order
needed, I just reused the cleanup from the tpp tests.

### DIFF
--- a/cohortextractor/cohortextractor.py
+++ b/cohortextractor/cohortextractor.py
@@ -746,7 +746,7 @@ def main():
     generate_codelist_report_parser.add_argument(
         "--output-dir",
         help="Location to store output files",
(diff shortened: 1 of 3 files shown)


bloodearnest:
This is a yak for https://github.com/opensafely-core/interactive.opensafely.org/issues/140

evansd (approved):
Looks good.

PR #845, opened 2022-07-15
fix: Variable date casting

Jongmassey:
Where date(time)-based variables are used as the bounds of a `BETWEEN` expression, cast them both to `date` whenever the reference field for the comparison is also cast to `date`

### DIFF
--- a/cohortextractor/date_expressions.py
+++ b/cohortextractor/date_expressions.py
@@ -318,7 +318,7 @@ class MSSQLDateFormatter(DateFormatter):
         if match:
             column_ref = match.group(1)
             if date_format == "YYYY-MM-DD":
-                return column_ref
+                return MSSQLDateFormatter.cast_as_date(column_ref)
             elif date_format == "YYYY-MM":
                 return self.date_function_first_day_of_month(column_ref)
             else:


cohortextractor/tpp_backend.py
```diff
@@ -653,6 +653,9 @@ def get_date_condition(self, table, date_expr, between):
         ]
         join_str = "\n".join(joins)
         if min_date_expr is not None and max_date_expr is not None:
+            if flags.WITH_END_DATE_FIX:
+                min_date_expr = MSSQLDateFormatter.cast_as_date(min_date_expr)
+                max_date_expr = MSSQLDateFormatter.cast_as_date(max_date_expr)
```
inglesp, line 658: Drive-by comment: do we also need to make this change on lines 664 and 666?

Jongmassey: quite possibly - we've only specifically investigated the behaviour wrt. `BETWEEN` but it stands to reason (and in my testing on a dummy table with mssql) that we should change those also

tests/test_tpp_backend.py
```diff
@@ -1714,6 +1714,100 @@ def test_patients_address_as_of():
...
+    study = StudyDefinition(
+        population=patients.all(),
+        event_count=patients.with_these_clinical_events(
+            foo_codes,
+            between=["2021-01-01", "2021-01-04"],
+            returning="number_of_matches_in_period",
+        ),
+    )
+    results = study.to_dicts()
+    event_count = 4 if flags.WITH_END_DATE_FIX else 2
```
evansd, line 1756: I think that rather than changing the test assertion based on the flag, it would be better to keep the test the same but say that it's expected to fail if the fix isn't in place. You can do this with `@pytest.mark.xfail`:
https://docs.pytest.org/en/7.1.x/how-to/skipping.html#condition-parameter

The test will then still show as failing, but the failure won't cause the whole suite to fail.

evansd: Just to explain a bit further: the thing which gave me pause here is that it just seems very odd to have tests which enforce known buggy behaviour. But it's quite common to have tests for desired behaviour that you known will currently fail, which is the purpose of the `xfail` marker.

It's not that I think anything bad will happen from the way things are currently done (especially given that the flag is presumably only temporary) it's just not very idomatic pytest so I thought it was worth mentioning.

evansd: As discussed, you've explained that you want to assert that the tests give a very specific wrong result when the flag is disabled, rather than just failing in any old way — so it makes sense to have this slightly unusual style of test in this slightly unusual situation!

evansd (approved):
This is great, thanks for wading through all this stuff!

As discussed, I think for the sake of belt-and-braces it would be worth changing this line here:
https://github.com/opensafely-core/cohort-extractor/blob/61fd344c16689a08dafb33fad225c584f740eb4a/cohortextractor/date_expressions.py#L321
so that it always returns a date:
```py
return self.cast_as_date(column_ref)
```

Even though, given your fix, this isn't actually necessary to address the specific bug, I still think we should do it. Otherwise this "shortcut" isn't actually semantically equivalent to the thing it's supposed to be shortcutting, and I can imagine that causing us problems elsewhere.

inglesp:
Thanks both!

PR #846, opened 2022-07-19
fix: Add warning to with_gp_consultations docs

inglesp:
This moves the warning from the bottom of the section to the top.

Addresses part of opensafely-core/opensafely-cli#120

image


CarolineMorton (approved):
Yes this looks good.

PR #851, opened 2022-07-22
fix: add SUS ethnicity mapping to docstring

### LINKED ISSUE #850: Ethnicity in SUS documentation
https://docs.opensafely.org/study-def-variables/#cohortextractor.patients.with_ethnicity_from_sus

> As a user I want to know which heading is which from the documentation and how the groupings are done so I can sense check and quality assure results.

Currently there is minimal information in the documentation here about what `group_6` and `group_16` might be. It makes it hard for a user to check particularly as we have multiple codelists for ethnicity now.

alexwalkerepi:
Based on formatting in #816.

Fixes #850

### DIFF
--- a/cohortextractor/patients.py
+++ b/cohortextractor/patients.py
@@ -2555,6 +2555,36 @@ def with_ethnicity_from_sus(
         return_expectations: a dictionary describing what dummy data should
             look like
 
+    Returns:
+        list: of integers, encoded (in at least TPP) as follows.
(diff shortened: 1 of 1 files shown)


iaindillingham (approved):

iaindillingham:
These changes won't appear in the documentation until we bump cohort-extractor and build the documentation. I think that happens daily.

PR #852, opened 2022-07-22
default to fixed date behaviour

Jongmassey:
Makes default the "date fix" (#845 ) whereby all `datetime`s are cast to `date`s in date comparisons. Previous behaviour (which is considered buggy) optionally available via `--without-end-date-fix`.


inglesp:
Sorry @Jongmassey, didn't see this when I added #865.

PR #853, opened 2022-08-08
Add defaults to with_these_medications docstring

Jongmassey:
Copy missing default and tie-breaking behaviour 
explanation for find_first_match_in_period and 
find_last_match_in_period args from 
with_these_clinical_events to
with_these_medications


iaindillingham (approved):
Thanks for documentng this. For archaeologists, the behaviour can be found in `TPPBackend._patients_with_events`.

PR #854, opened 2022-08-10
feat: Add with_at_least_one_day_in_critical_care filter to admitted_to_hospital

### LINKED ISSUE #828: Add with_at_least_one_day_in_critical_care filter to admitted_to_hospital
See [this Slack thread](https://bennettoxford.slack.com/archives/C02HJTL065A/p1655218771629069?thread_ts=1655217755.591279&cid=C02HJTL065A).

rebkwok:
Fixes #828

### DIFF
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -2460,6 +2460,7 @@ class TPPBackend:
         with_these_primary_diagnoses=None,
         with_these_diagnoses=None,
         with_these_procedures=None,
+        with_at_least_one_day_in_critical_care=False,
         **kwargs,
     ):
         supported_columns = {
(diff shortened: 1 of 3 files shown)


cohortextractor/tpp_backend.py
```diff
@@ -2548,6 +2549,9 @@ def patients_admitted_to_hospital(
             value_sql = ", ".join(map(quote, to_list(column_value)))
             conditions.append(f"{supported_columns[column_name]} IN ({value_sql})")
 
+        if with_at_least_one_day_in_critical_care:
+            conditions.append("APCS_Der.Spell_PbR_CC_Day > 0")
```
evansd, line 2553: This is a bit superfluous because MSSQL will spot that we're comparing to an int and CAST the string column automatically, but it might be clearer to do this explicitly e.g.
```suggestion
            conditions.append("CAST(APCS_Der.Spell_PbR_CC_Day AS int) > 0")
```

evansd (approved):
Looks good. One small suggestion which I'll leave to your judgement as to whether it's helpful or not.

cohortextractor/tpp_backend.py
rebkwok: Good point -  while I was adding the tpp_backend code I'd thought about that and meant to come back to it, but I was kind of expecting the tests to compain and remind me :)

PR #858, opened 2022-08-24
Add date range fix to EMIS

inglesp:


tests/emis_backend_setup.py
```diff
@@ -38,7 +39,24 @@
 from cohortextractor.trino_utils import wait_for_trino_to_be_ready
 from tests.helpers import mssql_sqlalchemy_engine_from_url, wait_for_mssql_to_be_ready
 
-Base = declarative_base()
+
+# Hack: in order to improve the variety of our test data, we intercept object
+# initialization, adding a time part to any kwarg that is for a datetime field that has
+# been provided as a date.
+class BaseTable:
```
Jongmassey, line 46: Not sure I understand the reasoning here

inglesp: It matches a change to [`tpp_backend_setup.py`](https://github.com/opensafely-core/cohort-extractor/blob/76eb09fd43d1bd42133dd537524d2120a0d5ac66/tests/tpp_backend_setup.py#L28-L44).

The rationale is the tests had all assumed that datetime fields only contained dates, which masked this bug.  So to improve the chance of flushing out any related issues, we should change all the dates to datetimes.

Jongmassey: I see, that makes perfect sense.

Jongmassey (approved):

PR #859, opened 2022-08-29
fix: Fix handling of NaN values

evansd:
These were getting introduced via the `patients.with_value_from_file()`
function when loading data saved using Pandas which encodes missing
values as `NaN`. This resulted in SQL like this:

    INSERT INTO #tmp4_vax_date_covid_1_file (patient_id, vax_date_covid_1) VALUES
    (123456, '20210101'),
    (123457, nan);

Which of course leads to the error:

    Invalid column name 'nan'

### DIFF
--- a/cohortextractor/emis_backend.py
+++ b/cohortextractor/emis_backend.py
@@ -1,4 +1,5 @@
 import datetime
+import math
 import os
 import re
 import uuid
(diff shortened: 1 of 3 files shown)


bloodearnest (approved):
Looks good.

PR #860, opened 2022-08-29
Support parameterising study definitions

evansd:
This teaches `generate_cohort` to accept a new, multi-valued `--param` argument e.g.

    cohortextractor generate_cohort --param key1=value1 --param key2=value2 --param key3

The study definition can access these via the `params` dict in the `cohortextractor` module:
```py
from cohortextractor import params
print(params)
# {'key1': 'value1', 'key2': 'value2', 'key3': ''}
```

There was a suggestion earlier of using `--arg` for this, but I think `--param` avoids confusion with other kinds of thing which are also called "arguments", and fits neatly with the term "parameterised study definition".

This was all much harder than it ought to be because implicit in this feature is the need for the output file name to be customised. Previously the output directory and format could be changed, but the file name was determined by the study definition name e.g. `study_definition_test.py` would always produce a file called `<output_dir>/input_test.<extension>`.

So we now support an `--output-file` argument which sets the directory, base name and file format in one go e.g. `some_dir/my_results.feather`.

### DIFF
--- a/cohortextractor/cohortextractor.py
+++ b/cohortextractor/cohortextractor.py
@@ -142,6 +142,7 @@ def generate_cohort(
     skip_existing=False,
     output_format=SUPPORTED_FILE_FORMATS[0],
     output_name=None,
+    params=(),
 ):
     preflight_generation_check()
     study_definitions = list_study_definitions()
@@ -174,6 +175,7 @@ def generate_cohort(
                 skip_existing=skip_existing,
                 output_format=output_format,
                 output_name=output_name or "input",
+                params=params,
             )
 
 
@@ -187,6 +189,7 @@ def _generate_cohort(
     skip_existing=False,
     output_format=SUPPORTED_FILE_FORMATS[0],
     output_name="input",
+    params=(),
 ):
     logger.info(
         f"Generating cohort for {study_name} in {output_dir}",
@@ -201,7 +204,7 @@ def _generate_cohort(
         output_format=output_format,
     )
 
-    study = load_study_definition(study_name)
+    study = load_study_definition(study_name, params=params)
 
     os.makedirs(output_dir, exist_ok=True)
 
@@ -562,20 +565,24 @@ def _make_cohort_report(input_dir, output_dir, study_name, suffix):
     logger.info(f"Created cohort report at {output_dir}/descriptives{suffix}.html")
 
 
-def dump_cohort_sql(study_definition):
-    study = load_study_definition(study_definition)
+def dump_cohort_sql(study_definition, params=()):
+    study = load_study_definition(study_definition, params=params)
     print(study.to_sql())
 
 
-def dump_study_yaml(study_definition):
-    study = load_study_definition(study_definition)
+def dump_study_yaml(study_definition, params=()):
+    study = load_study_definition(study_definition, params=params)
     print(yaml.dump(study.to_data()))
 
 
-def load_study_definition(name, value="study"):
+def load_study_definition(name, value="study", params=()):
     sys.path.extend([relative_dir(), os.path.join(relative_dir(), "analysis")])
     # Avoid creating __pycache__ files in the analysis directory
     sys.dont_write_bytecode = True
+    # Set any supplied "<key>=<value>" parameters on the global `params` dict
+    cohortextractor.params.clear()
+    cohortextractor.params.update(params)
+    logger.info(f"Setting cohortextractor.params to: {cohortextractor.params}")
     return getattr(importlib.import_module(name), value)
 
 
@@ -660,6 +667,14 @@ def main(args=None):
     dump_cohort_sql_parser.add_argument(
         "--study-definition", help="Study definition name", type=str, required=True
     )
+    dump_cohort_sql_parser.add_argument(
+        "--param",
+        nargs="+",
+        action="append",
+        dest="params",
+        default=[],
+        help="Additional parameter to pass to study definition",
+    )
     dump_cohort_sql_parser.set_defaults(which="dump_cohort_sql")
     dump_study_yaml_parser = subparsers.add_parser(
         "dump_study_yaml", help="Show study definition as YAML"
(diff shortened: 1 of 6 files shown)


tests/test_cohortextractor.py
```diff
@@ -32,3 +32,118 @@ def test_list_study_definition(relative_dir):
         relative_dir.return_value = dummy_repo
         definitions = list_study_definitions()
         assert definitions == [("study_definition_test", "_test")]
+
+
+class PatchStack(ExitStack):
+    """
+    Apply multiple `patch` context managers without nesting or using the ugly multi
+    argument form
```
rebkwok, line 40: :clap:

cohortextractor/cohortextractor.py
```diff
@@ -817,18 +859,37 @@ def main():
...
+                parser.error(
+                    "generate_cohort: error: cannot combine --output-file argument "
+                    "with --output-dir or --output-format"
+                )
+            output_dir = options.output_file.parent
+            match = re.match(
+                rf"^(.+)\.({EXTENSION_REGEX})$", str(options.output_file.name)
+            )
+            output_name = match.group(1)
+            output_format = match.group(2)
```
rebkwok, line 873: What happens if a `--output-file` has an unsupported extension?  Or some typo that means it doesn't match?  Does it need to be caught here to give a nicer error message than an error in the regex-match?

tests/fixtures/studies/test_params/analysis/study_definition.py
```diff
@@ -0,0 +1,24 @@
+from cohortextractor import StudyDefinition, params, patients
+
+age_suffix = params["age_suffix"]
+sex_suffix = params["sex_suffix"]
```
rebkwok, line 4: If an invalid parameter is passed to `generate_cohort`, will it get this far and raise a KeyError when the study def is loaded?  Will that get caught in a sensible way?

cohortextractor/cohortextractor.py
```diff
@@ -553,20 +565,23 @@ def _make_cohort_report(input_dir, output_dir, study_name, suffix):
...
 
 
-def load_study_definition(name, value="study"):
+def load_study_definition(name, value="study", params=()):
     sys.path.extend([relative_dir(), os.path.join(relative_dir(), "analysis")])
     # Avoid creating __pycache__ files in the analysis directory
     sys.dont_write_bytecode = True
+    # Set any supplied "<key>=<value>" parameters on the global `params` dict
+    cohortextractor.params.clear()
+    cohortextractor.params.update(param.partition("=")[::2] for param in params)
```
rebkwok, line 584: Would it be useful to log what the parameters are at this point?  If a user's done something weird, like `--param key,val` instead of `--param key=val`, the param will be interpreted as `{"key,val": ""}` and possibly not immediately obvious what they did wrong?

rebkwok (approved):
One typo and some questions that are all along the lines of "what happens if a user does this weird thing"; feel free to ignore if they're irrelevant

wjchulme:
This is fab! Couple of questions:

If I do `--param arg1 = 1` in the action and `arg1 = params["arg1"]` in the study def, is `arg1` a string `"1"` or a numeric `1`? What about `arg1 = True`? string or logical? I'm thinking of scenarios where we need the non-string type in the study definition, which might catch users out. For example I don't think `find_first_match_in_period = "True"` will work, so we'd need to do `arg1 = params["arg1"]=="True"` or something like that to get the correct type. 

Can I do `--param key1 = "string with white space"` ?

sebbacon:
🌊 🐦 : have you done documentation and "platform news" for this? Cos it's cool.

evansd:
Thanks @wjchulme. This is interesting feedback because it already exposes one misunderstanding which is that you can't write `key = value`, it has to be `key=value` because of the vagaries of bash's argument parsing (the former is considered to be three separate arguments, and would need to be quoted as `'key = value'` if you wanted it treated as a single one). I think command line folk get so used to this that they forget not everyone has the same form of Stockholm syndrome :)

That said, I think we can make the space separated version work with a bit of fiddling. We'd need to let `--param` accept an arbitrary number of arguments and then join them together before parsing them. And given that you won't be the only person to use spaces I think we'll have to support that as well.

You can definitely include whitespace. If we make the change above then your example would work exactly as written. Otherwise you'd have to write it as `--param 'key1=string with whitespace'`.

As for types, there isn't a neat solution with this approach. I think the only sane option is to make it clear that everything is a string and needs to be explicitly converted if you want something else.

There is a totally different approach we could take though that would give type support, and that's to use the `config:` option in `project.yaml`. That would let you write something like:
```yaml
generate_cohort:
  run: cohortextractor:latest generate_cohort --study-definition study_definition
  config:
    param_1: some string
    param_2: true
    param_3: 10
  outputs:
    highly_sensitive:
      cohort: output/input.csv
```

I actually think that would be less work than trying to add the support for spaces I described above. So maybe that's the way to go?

evansd:
> have you done documentation and "platform news" for this? Cos it's cool.

No, but was planning to do the docs once it was merged. And maybe the platform news after at least one person has tried it first?

wjchulme:
> everything is a string and needs to be explicitly converted if you want something else

This is how the equivalent functionality works in R, so at least there's some consistency. The problem is that most users won't know how to do the conversion in python. These are edges cases though -- I reckon a vast majority of parameters used will be needed as strings (even if conceptually they're numeric, eg `f"date + {param1} days"`).

wjchulme:
I think users would quickly get used to having to use `key=value`, not `key = value`. We could just advise that all key-value pairs should be expressed inside quotes to avoid problems. 

I do like the `config:` suggestion though. It's probably less error-prone. More sturdy.

evansd:
> I do like the config: suggestion though. It's probably less error-prone. More sturdy.

Yeah, I go a bit back and forth on this. It relies on coupling bits of the system together in ways which make me a bit uncomfortable. Not saying we definitely shouldn't do it, but I'd rather avoid baking that behaviour in at this stage.

For now I've just kept the `--param` syntax but implemented the whitespace handling.

cohortextractor/cohortextractor.py
evansd: Good point, I've added code to catch this now.

tests/fixtures/studies/test_params/analysis/study_definition.py
evansd: Also a good point. I've made `params` a custom dict subclass which raises a custom error in place of `KeyError` which is then caught by our usual error handling.

cohortextractor/cohortextractor.py
evansd: :+1: I've added logging now

PR #861, opened 2022-08-30
fix: Fix ambiguous column name

evansd:
If you try to return the matching category from a codelist using a date
drawn from another query which itself also returns a cateogry then you
get the error:

    Ambiguous column name 'category'

Example of a problematic query here:
https://github.com/opensafely/amr-uom-brit/blob/4c0879664ca/analysis/study_definition_ab_type_OS_new_2021.py#L110-L134


inglesp (approved):
Looks plausible!

Would it be easy to add a test?

evansd:
@inglesp It was. Thanks for the prompt!

PR #865, opened 2022-09-08
fix: Always enable date range fix

inglesp:
The deleted comment talks about adding a --without-fix flag, but I think that introduces unnecessary complexity and we can add it later if we need to.

### DIFF
--- a/cohortextractor/cohortextractor.py
+++ b/cohortextractor/cohortextractor.py
@@ -31,7 +31,6 @@ from pandas.api.types import (
 )
 
 import cohortextractor
-from cohortextractor import flags
 from cohortextractor.exceptions import DummyDataValidationError, ValidationError
 from cohortextractor.generate_codelist_report import generate_codelist_report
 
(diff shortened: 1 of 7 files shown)


cohortextractor/cohortextractor.py
```diff
@@ -738,10 +737,6 @@ def main(args=None):
         help="Do not regenerate data if output file already exists",
         action="store_true",
     )
-    generate_cohort_parser.add_argument(
-        "--with-end-date-fix",
-        action="store_true",
-    )
```
evansd, line 744: Is it worth leaving this in as a no-op, at least temporarily, so that studies using it don't break immediately. We could do a github search for people using it, get them all to remove it, and then remove the flag from here later.

evansd (approved):
Looks good. Just one question about making the migration path a bit more gentle.

cohortextractor/cohortextractor.py
inglesp: Very good point.  Have reinstated.

PR #873, opened 2022-10-12
fix: Handle another type of transient database error

evansd:
We've known about this type of error for ages (see [issue][1]) but when we first wrote this error handling code (#791) we hadn't seen it in a while so we decided to exclude it and see if it reoccurs. It has reoccurred; not terribly often but enough times that it's worth handling it here.

Note that transient errors which occur when downloading results are automatically handled by the retry code so we never see them. This error handling concerns errors which occur before we've finished generating the results.

[1]: https://github.com/opensafely-core/cohort-extractor/issues/776#issuecomment-1091675829

### DIFF
--- a/cohortextractor/cohortextractor.py
+++ b/cohortextractor/cohortextractor.py
@@ -913,7 +913,11 @@ def main(args=None):
 
             traceback.print_exc()
             # Exit with specific exit codes to help identify known issues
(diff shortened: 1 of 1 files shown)


bloodearnest (approved):
LGTM

It's given me an idea, but I"ll dump it in #tech

PR #875, opened 2022-10-20
fix: Support keywords as column names in `with_value_from_file`

evansd:


inglesp (approved):
Couldn't happen in databuilderland!

PR #887, opened 2022-11-24
fix: Don't double-count medications if codelist contains duplicate codes

rebkwok:


Jongmassey:
Could belt-and-braces it and be assured at the database level that there's no dupes in the codelist temptable by defining the code column as a PK in the codelist temptable creation which gets you a unique constraint and clustered index

```python
def create_codelist_table(self, codelist, case_sensitive=True):
        table_name = self.get_temp_table_name("codelist")
        if codelist.has_categories:
            values = list(codelist)
        else:
            values = list({(code, "") for code in codelist})
        # Depending on the case-sensitivity of the code system the columns in question
        # use different collations and we need to use a matching one here
        collation = "Latin1_General_BIN" if case_sensitive else "Latin1_General_CI_AS"
        max_code_len = max(len(code) for (code, category) in values)
        queries = [
            f"""
            -- Uploading codelist for {self._current_column_name}
            CREATE TABLE {table_name} (
              code VARCHAR({max_code_len}) COLLATE {collation} NOT NULL PRIMARY KEY,
              category VARCHAR(MAX)
            )
            """
        ]
        queries += make_batches_of_insert_statements(
            table_name, ("code", "category"), values
        )
        return table_name, queries
```


Although this
```python
if codelist.has_categories:
    values = list(codelist)
else:
    values = list({(code, "") for code in codelist})
```


Seems curious to me - if the codelist doesn't have categories it already gets deduped by going via a set, but not if it does have categories. Was there an intended use case for codes to be multiply-listed in different categories?

rebkwok:
> Seems curious to me - if the codelist doesn't have categories it already gets deduped by going via a set, but not if it does have categories. Was there an intended use case for codes to be multiply-listed in different categories?

No, codelists with categories should be similarly deduped.
Makes sense to add the PK as well (and avoids the need to manually make a clustered index on code after creating the table)

rebkwok:
@Jongmassey I've made those changes.  

Qn: Changing 
```
if codelist.has_categories:
    values = list(codelist)
```
to 
```
if codelist.has_categories:
    values = list(set(codelist))
```
assumes that we can never have a valid codelist where a code is in more than one category.  Is that correct (I _think_ it is)?

Jongmassey:
>  Is that correct (I think it is)?
It feels like it should be

evansd:
> assumes that we can never have a valid codelist where a code is in more than one category. Is that correct 

I think it's correct in these sense that any such codelist would be invalid, but I wouldn't want to guarantee that all the codelists we get will respect that property. So really to be safe we ought to be throwing an error if we get an inconsistent categorisation like this.

I wonder if we're better off doing that check in [`codelistlib.py`](https://github.com/opensafely-core/cohort-extractor/blob/main/cohortextractor/codelistlib.py) so the error gets thrown as soon as you try to load the codelist.

rebkwok:
@evansd I've added that consistency check in codelistlib.  I could make it also dedupe codes there, but I haven't (mostly because it would involve fixing up more tests), so it'll allow duplicate codes and (code, category) tuples, but dedupe them in the backends.

evansd:
@rebkwok Great, thanks. That sounds reasonable to me.

PR #896, opened 2022-12-13
Handle changing vmp ids

inglesp:


cohortextractor/codelistlib.py
```diff
@@ -77,3 +77,31 @@ def combine_codelists(first_codelist, *other_codelists):
...
+                code_to_category[id] = code_to_category[prev_id]
+        return codelist(sorted(code_to_category.items()), cl.system)
+    else:
+        codes = set(cl)
+        for id, prev_id in vmp_mapping:
+            if id in codes:
+                codes.add(prev_id)
+            if prev_id in codes:
+                codes.add(id)
+        return codelist(sorted(codes), cl.system)
```
rebkwok, line 107: if `TPPBackend.vmp_mapping` built a dict of `{id: set(other_ids)}` instead of a list of tuples, I think this could avoid checking every mapping tuple and just be:
```
codes_with_alternative_ids = codes & set(vmp_mapping)
for code in codes_with_alternative_ids:
    codes.update(vmp_mapping[code])
```
It would mean doing more work in the `TPPBackend.vmp_mapping` to make a dict that has all previous and subsequent ids for each code in the table, but since that's a cached property, maybe it's worth it?

In fact, I think it would just be one loop of the tuples there, rather than here for every codelist, i.e.:
```
mapping_dict = {}
for current, previous in vpm_mapping:
    mapping_dict.setdefault(current, set()).add(previous)
    mapping_dict.setdefault(previous, set()).add(current)
```

rebkwok (approved):
LGTM, one suggestion for the codelist expanding

evansd:
One very basic and naive question here: where does the `VmpMapping` table come from? Apologies if this has already been covered at length elsewhere.

One thought I had was that if it's being written by us then it's probably going to end up in the temporary database as we don't have write access to the main one.

inglesp:
> One very basic and naive question here: where does the `VmpMapping` table come from? Apologies if this has already been covered at length elsewhere.
> 
> One thought I had was that if it's being written by us then it's probably going to end up in the temporary database as we don't have write access to the main one.

I've not written the code to create the `VmpMapping` table yet!  But you're right about the temporary database.

However, I'm not sure about how to do the plumbing here.  Outside of tests, will `TPPBackend.temporary_database` always be set?

inglesp:
@evansd does https://github.com/opensafely-core/cohort-extractor/pull/896/commits/e79ee5abf1842c01540c7f07ddeac74211aa15bf address your concerns?

evansd (approved):
Yes, it does. I think the conditional testing here is fine.

cohortextractor/codelistlib.py
inglesp: Sorry, missed this when you wrote it.  It looks like a sensible suggestion, but I don't think there's any need to make this more efficient, as it's already much faster than talking to the database.

PR #898, opened 2022-12-16
feat: Add update_vmp_mapping command

inglesp:
This command updates the VmpMapping table with the latest data from OpenCodelists, accessed via the proxy.

### DIFF
--- a/cohortextractor/update_vmp_mapping.py
+++ b/cohortextractor/update_vmp_mapping.py
@@ -1,0 +1,42 @@
+import os
+
+import requests
+
+from .mssql_utils import mssql_dbapi_connection_from_url
+
+
+def update_vmp_mapping():
+    """Update VmpMapping table with latest data from OpenCodelists"""
+
+    mapping = [tuple(record) for record in get_vmp_mapping()]
+    temporary_database = os.environ["TEMP_DATABASE_NAME"]
+    conn = mssql_dbapi_connection_from_url(os.environ["DATABASE_URL"])
+    # Transaction behaviour cribbed from
+    # cohortextractor.tpp_backend.save_results_to_temporary_db
+    previous_autocommit = conn.autocommit_state
+    conn.autocommit(False)
+    cursor = conn.cursor()
+    cursor.execute(
+        f"""
+        IF OBJECT_ID('{temporary_database}..VmpMapping', 'U') IS NOT NULL
+            DROP TABLE {temporary_database}..VmpMapping
+
+        CREATE TABLE {temporary_database}..VmpMapping (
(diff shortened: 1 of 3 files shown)


cohortextractor/update_vmp_mapping.py
```diff
@@ -0,0 +1,36 @@
...
+def update_vmp_mapping():
+    """Update VmpMapping table with latest data from OpenCodelists"""
+
+    mapping = [tuple(record) for record in get_vmp_mapping()]
+    temporary_database = os.environ["TEMP_DATABASE_NAME"]
+    conn = mssql_dbapi_connection_from_url(os.environ["DATABASE_URL"])
+    cursor = conn.cursor()
+    cursor.execute(
+        f"""
+IF OBJECT_ID('{temporary_database}..VmpMapping', 'U') IS NOT NULL
```
evansd, line 17: quecto-nit: I'd find this slightly easier to read if the SQL was indented to the level of the closing quotes. The database won't care about the extra whitespace, and it's what we do with inline SQL elsewhere in cohortextractor.

cohortextractor/update_vmp_mapping.py
```diff
@@ -0,0 +1,36 @@
...
+    id VARCHAR(18) COLLATE Latin1_General_CI_AS,
+    prev_id VARCHAR(18) COLLATE Latin1_General_CI_AS,
+)
+        """
+    )
+    cursor.executemany(
+        f"INSERT INTO {temporary_database}..VmpMapping VALUES (%s, %s)",
+        mapping,
+    )
+    conn.commit()
```
evansd, line 30: I assume the intention here is that all this happens in a transaction so the DROP/re-CREATE doesn't disrupt any running queries. Are you confident that's happening here? I remember having to do some dead chicken waving elsewhere to get a proper transaction e.g.
https://github.com/opensafely-core/cohort-extractor/blob/9da8873acbc473f16e536cb241fdadb6e98a0e8e/cohortextractor/tpp_backend.py#L226-L237

You could check by sticking some bad values in `mapping` so that the INSERT blows up and then checking to see if the original table is still there.

cohortextractor/update_vmp_mapping.py
```diff
@@ -0,0 +1,36 @@
...
+    conn.close()
+
+
+def get_vmp_mapping():
+    rsp = requests.get("https://opencodelists-proxy.opensafely.org/api/v1/dmd-mapping/")
```
evansd, line 35: The `.json()` call is probably sufficient error checking here, but out of paranoia I generally like to call `rsp.raise_for_status()` after each request to make sure I'm not working with an error response.

evansd (approved):

Issue #856, opened 2022-08-16
EMIS Starburst image has updated to 391-e


StevenMaude:
[Automated check failure](https://github.com/opensafely-core/cohort-extractor/runs/7853388986?check_suite_focus=true):

![A screenshot showing a test failure in GitHub Actions, with a different version of the Starburst image run by EMIS to that in this repository.](https://user-images.githubusercontent.com/3818079/184855289-622cada6-b58c-4514-a79c-42d0bed91530.png)

I tried updating, but [there are other test failures](https://github.com/opensafely-core/cohort-extractor/actions/runs/2859385635) with errors like:

```
PrestoExternalError(type=EXTERNAL, name=JDBC_ERROR, message="The driver could not establish a secure connection to SQL Server by using Secure Sockets Layer (SSL) encryption. Error: "PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target". ClientConnectionId:1272c9a2-baae-4bd0-858f-5aca27eecbd1", query_id=20220816_082956_00059_5zqk3)
```

These apparently are due to the database not being connectable, so it might be:

* the client needs updating, and/or
* there's some other configuration change that's needed

StevenMaude:
We should then re-enable the scheduled check disabled in #905, perhaps with lower frequency.

PR #905, opened 2023-01-04
Temporarily remove scheduled Trino version check

StevenMaude:
#856 has been open for a few months and it has not been resolved.

In the meantime, this check only serves to spam my GitHub notifications every single weekday, to inform of the same problem.

This change should be reverted when #856 is fixed.


StevenMaude:
An alternative could be reduce the frequency to once per month, if disabling entirely isn't desirable.

inglesp (approved):

PR #922, opened 2023-02-14
fix: Switch ONS CIS `visit_id` to `pseudo_visit_id`

iaindillingham:
`ONS_CIS.visit_id` has become `ONS_CIS_New.pseudo_visit_id`, and has switched from a `varchar` to a `varbinary`. We use this column to break ties, and don't reveal it to the user, so we simply switch names and types.

### DIFF
--- a/cohortextractor/ons_cis_utils.py
+++ b/cohortextractor/ons_cis_utils.py
@@ -79,7 +79,7 @@ ONS_CIS_COLUMN_MAPPINGS = {
     "travel_abroad": "bool",
     "travel_abroad_date": "date",
     "visit_date": "date",
-    "visit_id": "str",
+    "pseudo_visit_id": "bytes",
     "work_direct_contact_patients_etc": "bool",
     "work_location": "int",
     "work_outside_home_days": "int",
--- a/cohortextractor/study_definition.py
+++ b/cohortextractor/study_definition.py
@@ -291,6 +291,8 @@ class StudyDefinition:
                 dtypes[name] = "category"
             elif column_type == "float":
                 dtypes[name] = "float"
+            elif column_type == "bytes":
+                dtypes[name] = "bytes"
             else:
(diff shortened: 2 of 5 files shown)


cohortextractor/ons_cis_utils.py
```diff
@@ -79,7 +79,7 @@
     "travel_abroad": "bool",
     "travel_abroad_date": "date",
     "visit_date": "date",
-    "visit_id": "str",
+    "pseudo_visit_id": "bytes",
```
rebkwok, line 82: Does this definitely work?  I remember not being able to add certain types here because it causes pandas errors in the converting for output

iaindillingham: We don't extract `pseudo_visit_id`, so I don't think Pandas will error; I hope the tests would tell me, if not. Did you have a test in mind?

rebkwok: Actually, I expect it does in the current tests, because it's only being used for tie-breaking.  I think it will break if you try to return `pseudo_visit_id` in the output (e.g. in `tests/test_tpp_backend.py:test_ons_cis`)

iaindillingham:
Would you mind taking a second look, @rebkwok? As you suggested, I've returned `pseudo_visit_id` in the output:

https://github.com/opensafely-core/cohort-extractor/blob/5652352bac7825b8e66ffbc41c12733851077939/tests/test_tpp_backend.py#L6551-L6555

You'll notice a `FIXME`. I don't understand why the default value for `"bytes"` shouldn't be `bytes([0])`. Could you help?

https://github.com/opensafely-core/cohort-extractor/blob/5652352bac7825b8e66ffbc41c12733851077939/cohortextractor/tpp_backend.py#L440-L455

rebkwok (approved):

PR #961, opened 2023-06-23
Add `update_custom_medication_dictionary`

iaindillingham:
This sub-command updates the `CustomMedicationDictionary` table. (In reality, it first drops and then creates it, but it's called "update" for consistency with `update_vmp_mapping`.) This table is required by opensafely-core/ehrql#1270. Whilst its maintenance should, arguably, be handled by a separate tool, we already drop and create tables in the temporary database in cohort-extractor, so the maintenance of this table is here for consistency.

The implementation follows that of `cohortextractor.update_vmp_mapping` closely. I've added a couple of QOL improvements to the test and, to preserve my sanity, avoided using `unittest.mock`.

### DIFF
--- a/cohortextractor/custom_medication_dictionary.csv
+++ b/cohortextractor/custom_medication_dictionary.csv
@@ -1,0 +1,1 @@
+DMD_ID,MultilexDrug_ID
--- a/cohortextractor/cohortextractor.py
+++ b/cohortextractor/cohortextractor.py
@@ -32,6 +32,9 @@ from pandas.api.types import (
 
 import cohortextractor
 from cohortextractor.exceptions import DummyDataValidationError, ValidationError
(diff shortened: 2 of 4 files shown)


cohortextractor/custom_medication_dictionary.csv
```diff
@@ -0,0 +1 @@
+DMD_ID,MultilexDrug_ID
```
iaindillingham, line 1: @inglesp, could you tell me where I might find the mappings?

evansd (approved):
Looks good to me!

You're right that DROP/CREATE isn't strictly an update. But given that it all happens inside a transaction I think it's observably indistinguishable from an update by the rest of the system, so I don't think it's completely misnamed.

PR #964, opened 2023-06-28
Update the custom medication dictionary

iaindillingham:
Thanks for providing the mappings, @chrisjwood16. For background, see [1]; and opensafely-core/ehrql#1270.

[1]: https://bennettoxford.slack.com/archives/C03HVM72273/p1687786350885009

### DIFF
--- a/cohortextractor/custom_medication_dictionary.csv
+++ b/cohortextractor/custom_medication_dictionary.csv
@@ -1,1 +1,4 @@
-DMD_ID,MultilexDrug_ID
+DMD_ID,MultilexDrug_ID,FullName
+36128211000001109,14319;1;0,morphine sulfate injection 10mg/1ml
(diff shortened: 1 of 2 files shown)


inglesp (approved):
Is it worth having another column for the drug name?

iaindillingham:
> Is it worth having another column for the drug name?

Do you mean to the CSV and the table? Or just to the CSV? I've added it to the CSV, and updated the tests, because I wasn't sure that adding an extra column to the CSV wouldn't break the `INSERT`. (It doesn't!) However, I haven't added it to the table. (Possibly this should have featured in opensafely-core/ehrql#1348?)

inglesp:
> Do you mean to the CSV and the table? Or just to the CSV?

Just the CSV!

PR #968, opened 2023-07-07
Avoid locking sys.table when creating temporary tables

### LINKED ISSUE #965: Avoid locking `sys.tables` when using `SELECT ... INTO`
cohort-extractor writes results to a table in the temporary database (`OpenCORONATempTables`) using `SELECT ... INTO`

https://github.com/opensafely-core/cohort-extractor/blob/85a92b06adc5c13e85637953ca0a81eb1012b6a6/cohortextractor/tpp_backend.py#L65-L80

However, [`SELECT ... INTO` can lock `sys.tables` for long periods of time][1]; in effect, cohort-extractor jobs run in sequence.[^1]

To avoid locking `sys.tables` for long periods of time, we could first create the table in the temporary database and then, in a separate transaction, populate it. We would, however, need to be able to determine its schema. We might, for example:

```sql
SELECT TOP(0) INTO OpenCORONATempTables..DataExtract_<hash> FROM (...)
```

@evansd and @Jongmassey discuss this issue in [this Slack thread][2].

[^1]: Indeed, any jobs that modify the database run in sequence.

[1]: https://www.sqlshack.com/sql-server-lock-issues-when-using-a-ddl-including-select-into-clause-in-long-running-transactions/
[2]: https://bennettoxford.slack.com/archives/C63UXGB8E/p1687962434219019

evansd:
Uses the same approach described in:

 * https://github.com/opensafely-core/ehrql/pull/1400

Closes #965

### DIFF
--- a/cohortextractor/tpp_backend.py
+++ b/cohortextractor/tpp_backend.py
@@ -230,7 +230,17 @@ class TPPBackend:
             logger.info(f"Writing results into temporary table '{output_table}'")
             conn = self.get_db_connection()
             cursor = conn.cursor()
-            cursor.execute(f"SELECT * INTO {output_table} FROM ({final_query}) t")
+            # Use a `SELECT INTO` with a WHERE clause that will never match to create a
+            # temporary table with the right structure without needing to fetch all the
+            # results. (Creating a table takes a lock on `sys.tables` so we want this to
+            # be as fast as possible.)
+            cursor.execute(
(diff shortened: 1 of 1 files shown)


inglesp (approved):
Wasn't expecting this to have a net negative number of lines!

evansd:
Ha! I hadn't spotted that. Yes, that is unexpected.

Have run a similar test to [this one](https://github.com/opensafely-core/ehrql/pull/1400#issuecomment-1625406966) to confirm that this works in real life.

Issue #955, opened 2023-05-23
`not-from-a-package` version is not PEP 440 compliant, breaking `pip-compile`


StevenMaude:
When trying to rebuild the requirements:

```
$ pip-compile --extra=drivers --output-file=requirements.prod.txt setup.py
Backend subprocess exited when trying to invoke get_requires_for_build_wheel
Failed to parse …cohort-extractor/setup.py
```

This isn't a particularly helpful error (see jazzband/pip-tools#1711).

If you get information on the underlying subprocess error, then you see this warning from `setuptools`:

```
SetuptoolsDeprecationWarning: Invalid version: \'not-from-a-package\'.

!!

********************************************************************************
The version specified is not a valid version according to PEP 440.
This may not work as expected with newer versions of
setuptools, pip, and PyPI.

By 2023-Sep-26, you need to update your project and remove deprecated calls\n
or your builds will no longer be supported.

See https://peps.python.org/pep-0440/ for details.
```

However the build actually does fail, even though we're not in September 2023 yet, due to the `packaging` package, vendored by `setuptools`; you get the following error that causes the `pip-compile` failure:

```
…
File "…/site-packages/setuptools/_vendor/packaging/version.py", line 197, in __init__
raise InvalidVersion(f"Invalid version: \'{version}\'")
setuptools.extern.packaging.version.InvalidVersion: Invalid version: \'not-from-a-package\'
```

This is with the existing provided requirements:

```
$ pip freeze # minimal virtualenv created with only: pip install pip-tools; with the pyproject_hooks package not listed in the requirements uninstalled
build==0.8.0
click==8.1.0
packaging==23.0
pep517==0.12.0
pip-tools==6.13.0
tomli==1.2.3
$ cat requirements.dev.txt | grep -E "^build|^click|^packaging|^pep517|^pip-tools|^tomli"
build==0.8.0
click==8.1.0
packaging==23.0
pep517==0.12.0
pip-tools==6.13.0
tomli==1.2.3
```

Issue #895, opened 2022-12-13
Dependabot doesn't seem to ever update `requirements.prod.txt`


StevenMaude:
All of Dependabot's [automatic dependency updates](https://github.com/opensafely-core/cohort-extractor/pulls?q=is%3Apr+is%3Aclosed+label%3Adependencies) are either for the `requirements.dev.txt` or actions used in GitHub Actions workflows.

PR #979, opened 2023-08-10
Replace `Makefile` with `justfile`

StevenMaude:
And do an overhaul of the repository structure to use Ruff.


StevenMaude:
Issues:
* Need to fix the Docker build.
* The just recipes to install the dev and production requirements have some duplication; there might be a nicer way of passing the `--extra=drivers` arguments, *and* other extra arguments. Just's documentation is not easy to find this.
* Not for this PR, but we still have issues with updating the dependencies:
  * #955 (this can be temporarily worked around by setting a valid version locally)
  * #895, which is some conflicting version issue

justfile
```diff
@@ -0,0 +1,148 @@
...
+
+# build the Docker image
+docker-build env="dev" *args="":
+    export BUILD_DATE=$(shell date +'%y-%m-%dT%H:%M:%S.%3NZ')
+    export REVISION=$(shell git rev-parse --short HEAD)
+    # We retrieve the version in GitHub Actions differently,
+    # so allow overriding this value.
+    export VERSION=${VERSION:-$(shell git describe --tags)}
+    export DOCKER_BUILDKIT=1
+    docker-compose build --pull {{ args }} {{ env }}
```
madwort, line 120: I think you need to use `#!/usr/bin/env bash` for these env vars to be picked up by docker-compose - the just default is that each line is run in a separate shell, so they will be lost. e.g.

```
# justfile
myenv:
  export MYENVVAR=yeahyeahyeah
  echo $MYENVVAR
```

```bash
% just myenv
export MYENVVAR=yeahyeahyeah
echo $MYENVVAR
sh: MYENVVAR: unbound variable
error: Recipe `myenv` failed on line 25 with exit code 1
```

madwort (changes requested):
I don't think `just docker-build` will work as expected

justfile
```diff
@@ -0,0 +1,148 @@
...
+
+
+# update requirements.prod.txt if requirements.prod.in has changed
+requirements-prod *args:
+    "{{ just_executable() }}" _compile-prod setup.py requirements.prod.txt {{ args }}
+
+
+# update requirements.dev.txt if requirements.dev.in has changed
+requirements-dev *args: requirements-prod
+    "{{ just_executable() }}" _compile-dev requirements.dev.in requirements.dev.txt {{ args }}
```
madwort, line 63: I think we could condense this like so:
```suggestion
_compile src dst env *args: virtualenv
    #!/usr/bin/env bash
    set -euo pipefail

    # exit if src file is older than dst file (-nt = 'newer than', but we negate with || to avoid error exit code)
    test "${FORCE:-}" = "true" -o {{ src }} -nt {{ dst }} || exit 0
    $BIN/pip-compile --allow-unsafe {{ if env == "dev" { "--extras=drivers" } else { "" } }} --output-file={{ dst }} {{ src }} {{ args }}


# update requirements.prod.txt if requirements.prod.in has changed
requirements-prod *args:
    "{{ just_executable() }}" _compile-prod setup.py requirements.prod.txt prod {{ args }}


# update requirements.dev.txt if requirements.dev.in has changed
requirements-dev *args: requirements-prod
    "{{ just_executable() }}" _compile-dev requirements.dev.in requirements.dev.txt dev {{ args }}
```

justfile
```diff
@@ -0,0 +1,148 @@
+# just has no idiom for setting a default value for an environment variable
+# so we shell out, as we need VIRTUAL_ENV in the justfile environment
+export VIRTUAL_ENV  := `echo ${VIRTUAL_ENV:-.venv}`
```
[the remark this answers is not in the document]
StevenMaude: Maybe we should update it in the repo template?

https://github.com/opensafely-core/repo-template/blob/e3d84fe3981d7f3df8f3cfa3c3a0a72e33cea5f2/justfile#L3

justfile
StevenMaude: Thanks! I wasn't sure what the idiomatic way of doing things is.

(Just's README is very long, and I didn't find it easy to find much in its documentation either.)

madwort: tbh I'm still mostly figuring out what's possible in justfiles, more than what's idomatic or conventional. I found the [just programmer's manual](https://just.systems/man/en/chapter_1.html) hit the right spot for me in terms of being able to grok some useful features.

actually now you mention it, I think we can make this probably more idiomatic by removing the use of `just_executable()` & making `requirements-*` only have dependencies, like this:

```
# update requirements.dev.txt if requirements.dev.in has changed
requirements-dev *args: requirements-prod (_compile-dev "requirements.dev.in" "requirements.dev.txt" "dev" args)
```
(not sure if this is much better or not, but kind of nice not to invoke a child just process...)

justfile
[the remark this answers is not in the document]
madwort: we probably should...! I notice that was written two years ago - it's very possible that `just` did not have this feature when that comment was written...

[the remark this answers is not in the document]
madwort: (do you want to update the template, or should I?)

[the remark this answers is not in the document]
madwort: thinking about it - it would be good to land these changes here & test them before we port it upstream to the template?

justfile
iaindillingham, line 1: If I run `just devenv` in a new shell, then I see:

```
﻿Backend subprocess exited when trying to invoke get_requires_for_build_wheel
Failed to parse /Users/iaindillingham/Code/opensafely-core/cohort-extractor/setup.py
error: Recipe `_compile-prod` failed with exit code 2
error: Recipe `requirements-prod` failed on line 58 with exit code 2
```

If I run `just virtualenv`, `source .venv/bin/activate`, and `pip install .`, then within the stack trace I see:

```
Version 'not-from-a-package' is not valid according to PEP 440.
```

Admittedly, I see the same when I run `make devenv`; but I think we should fix this issue before replacing the makefile with the justfile.

madwort: We should fix this, but I think you're already seeing the entire output from `pip-compile`, so I don't think there's a difference between make & just - maybe a separate ticket? 

```
cohort-extractor % ./.venv/bin/pip-compile --allow-unsafe --extra=drivers --output-file=requirements.prod.txt setup.py
Backend subprocess exited when trying to invoke get_requires_for_build_wheel
Failed to parse /Users/tom/Documents/bennett-institute/cohort-extractor/setup.py
```

iaindillingham: True. I've made #980. It blocks my review, I'm afraid.

benbc:
@StevenMaude @inglesp is it worth doing this given that Cohort Extractor is now deprecated? Or should we just close this PR?

